PR #57, opened 2015-01-07
Fixes default rake task

deivid-rodriguez:
When I try to add the `haml-lint` rake task to my `Rakefile`, I get the following error: `File path './**/*.haml' does not exist`. I'm doing the "standard":

    require 'haml_lint/rake_task'
    HamlLint::RakeTask.new

This commit should fix it. I added a test borrowed from `scss-lint`.

### DIFF
--- a/lib/haml_lint/rake_task.rb
+++ b/lib/haml_lint/rake_task.rb
@@ -32,7 +32,7 @@ module HamlLint
       @exclude_linter = []
       @config = ''
       @exclude = []
-      @pattern = %w[./**/*.haml]
+      @pattern = Dir.glob('./**/*.haml')
     end
 
     def cli_args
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I've fixed travis failures. The workaround to this failures also thought me a workaround for my issue:

    require 'haml_lint/rake_task'
    HamlLint::RakeTask.new.pattern = Dir.glob('app/views/**/*.haml')

sds:
Thanks for the fix, @deivid-rodriguez.

Having had a closer look at this rake task, it didn't seem to align well with the conventions used by the upstream `rake` task libraries themselves, so I've made some breaking changes to it in 271d4e983fa2986cc38ec2041679c71c79c954a0, though I don't think it will affect many.

deivid-rodriguez:
Awesome improvements! Thanks a lot!!

PR #58, opened 2015-01-10
Add explicit color option

danielma:
I like running haml-lint in Guard, but Guard always returns `@out.tty?` as `false`, so I followed RSpec's lead and added an explicit --color option

### DIFF
--- a/lib/haml_lint/cli.rb
+++ b/lib/haml_lint/cli.rb
@@ -40,6 +40,7 @@ module HamlLint
     attr_reader :log
 
     def act_on_options(options)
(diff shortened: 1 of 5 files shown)


sds:
Thanks for the pull request, @danielma!

I've merged your change in 408d0a7 with some adjustments in bfeb66c. The main issue I had was that the implementation would not respect the `--no-color` flag for TTYs since it would default to enabling color due to the output stream being a TTY.

danielma:
@sds awesome! glad I could help out a little bit

PR #60, opened 2015-01-15
Fix line number color for light background ttys

deivid-rodriguez:
Currently the line number is almost invisible with light background
screens. I think using the same color as the file looks fine and it's a
good compromise.

Another option would be chosing another color that looks fine in all terminals, but I went for the easiest.

### DIFF
--- a/lib/haml_lint/reporter/default_reporter.rb
+++ b/lib/haml_lint/reporter/default_reporter.rb
@@ -17,7 +17,7 @@ module HamlLint
     def print_location(lint)
       log.info lint.filename, false
       log.log ':', false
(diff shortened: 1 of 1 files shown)


sds:
Thanks for the pull request, @deivid-rodriguez.

The solution to the problem is to instead not explicitly output the text as white (but still have it be bold). This has the same visual effect on dark backgrounds as before, but results in dark bolded text on light backgrounds. I've done this in 9df12f7ab8b5eaf714617ab593dbc8998bd1c4b0.

Thanks for the report!

deivid-rodriguez:
It works! Thanks a lot!

PR #63, opened 2015-02-10
[WIP] Reporter with RuboCop-style JSON output

### LINKED ISSUE #62: Rubocop-like JSON output
Hi!

I would like ask that are you planning to add an optional rubocop-like json output format?
We have a [bot](https://github.com/ManageIQ/miq_bot), which checks pull requests with rubocop.
I would like to add support for the `haml` files with haml-lint, but I need JSON as output.

Thanks!

skateman:
Hopefully resolves #62, but it needs further testing and some specs are still missing.

### DIFF
--- a/lib/haml_lint/cli.rb
+++ b/lib/haml_lint/cli.rb
@@ -40,6 +40,9 @@ module HamlLint
       elsif options[:show_linters]
         print_available_linters
         Sysexits::EX_OK
+      elsif options[:show_reporters]
+        print_available_reporters
+        Sysexits::EX_OK
       else
         scan_for_lints(options)
       end
@@ -86,6 +89,18 @@ module HamlLint
 
       linter_names.sort.each do |linter_name|
         log.log " - #{linter_name}"
+      end
+    end
+
+    def print_available_reporters
+      log.info 'Available reporters:'
+
+      reporter_names = Reporter.descendants.map do |reporter|
+        reporter.name.split('::').last.sub(/Reporter$/, '').downcase
+      end
+
+      reporter_names.sort.each do |reporter_name|
+        log.log " - #{reporter_name}"
       end
     end
 
--- a/lib/haml_lint/reporter.rb
+++ b/lib/haml_lint/reporter.rb
@@ -18,6 +18,15 @@ module HamlLint
       raise NotImplementedError
     end
 
+    # Keep tracking all the descendants of this class for the list of available reporters
+    def self.descendants
+      @descendants ||= []
+    end
+
+    def self.inherited(descendant)
+      descendants << descendant
+    end
+
     private
 
     attr_reader :log
--- a/lib/haml_lint/reporter.rb
+++ b/lib/haml_lint/reporter.rb
@@ -5,12 +5,14 @@ module HamlLint
   # @abstract
   class Reporter
     attr_reader :lints
+    attr_reader :files
 
     # @param logger [HamlLint::Logger]
     # @param report [HamlLint::Report]
     def initialize(logger, report)
       @log = logger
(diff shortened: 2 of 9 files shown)


lib/haml_lint/cli.rb
```diff
@@ -67,13 +70,14 @@ def handle_exception(ex)
     end
 
     def scan_for_lints(options)
-      report = Runner.new.run(options)
-      print_report(report, options)
+      runner = Runner.new
+      report = runner.run(options)
+      print_report(report, options, runner.files)
```
sds: You aren't using `runner.files` anywhere, so this modification is unnecessary.

lib/haml_lint/reporter.rb
```diff
@@ -5,12 +5,15 @@ module HamlLint
...
+    attr_reader :files
 
     # @param logger [HamlLint::Logger]
     # @param report [HamlLint::Report]
-    def initialize(logger, report)
+    # @param files  [Array]
+    def initialize(logger, report, files)
       @log = logger
       @lints = report.lints
+      @files = files
```
sds: This never gets used from what I can see (you can always extract the list of relevant files from the list of `lints`). It can safely be removed.

lib/haml_lint/reporter/json_reporter.rb
```diff
@@ -0,0 +1,34 @@
...
+          :hamllint_version => VERSION,
+          :ruby_engine      => RUBY_ENGINE,
+          :ruby_patchlevel  => RUBY_PATCHLEVEL.to_s,
+          :ruby_platform    => RUBY_PLATFORM
+        },
+        :files    => grouped.map { |l| map_file(l) },
+        :summary  => {
+          :offense_count        => lints.length,
+          :target_file_count    => grouped.length,
+          :inspected_file_count => files.length
```
sds: Sorry, I missed this tiny use of `files`, so my other comments aren't quite valid.

How valuable do you think this information is?

lib/haml_lint/reporter.rb
skateman: It's being used [here](https://github.com/skateman/haml-lint/commit/10d89c63581dc4d9e3001370daeb909a66e9e1b5#diff-202b23a957ac37cd746eb7ec9840c680R17), but it's in a separate commit. The `lints` shows only the files with issues if I'm right, but correct me if not!

lib/haml_lint/reporter/json_reporter.rb
skateman: @sds, sccording to the [RuboCop specification](http://www.rubydoc.info/gems/rubocop/0.11.0#JSON_Formatter), the summary section of the JSON output has the number of the inspected files. I didn't wanted to pass just the number, because this way it's possible to create e.g. a "positive-reporter", which shows you the files without any offenses.

lib/haml_lint/cli.rb
skateman: See below :)

sds:
Thanks for answering my questions quickly, @skateman.

Looks good, but would love to see some tests of course.

I would also like to see the `files` passed directly to the `Report` class, as it makes sense that the report would contain the list of files that were scanned. That would save you from having to modify the interface of the `Runner` constructor, or anything in the `CLI` class really. The reporter could then access the files directly from the `Report`.

Thanks for your work on this so far!

skateman:
That's a good idea, I'm on it!

skateman:
@sds I've fixed it and added some tests too, is it enough?

sds:
Merged in 7d04364. Thanks!

skateman:
Thanks!

PR #70, opened 2015-03-13
Check for missing alt attributes on img tags

### LINKED ISSUE #69: accessibility rules - check for missing alt attributes
Have you considered adding some basic accessibility assertions to your rules (i.e. presence of `alt` attributes, `label` association on form elements)?

I'd like to work on some PRs for accessibility rules, starting with looking for missing `alt` attributes. Would you be open to that?

ckundo:
This introduces a check that aligns with the W3C WCAG 2.0 guideline 1.1.1, which states:

> All non-text content that is presented to the user has a text alternative that serves the equivalent purpose

http://www.w3.org/TR/2008/REC-WCAG20-20081211/#text-equiv-all

[closes #69]

### DIFF
--- a/lib/haml_lint/linter/alt_text.rb
+++ b/lib/haml_lint/linter/alt_text.rb
@@ -1,0 +1,11 @@
+module HamlLint
+  class Linter::AltText < Linter
+    include LinterRegistry
+
+    def visit_tag(node)
+      if node.tag_name == 'img' && !node.has_hash_attribute?(:alt)
+        add_lint(node, "`img` tags must include alt text.")
+      end
+    end
+  end
+end
--- a/lib/haml_lint/tree/tag_node.rb
+++ b/lib/haml_lint/tree/tag_node.rb
@@ -20,6 +20,14 @@ module HamlLint::Tree
       @value[:parse] && !@value[:value].strip.empty?
     end
 
+    # Returns whether this tag has a specified attribute
+    # `=`).
+    #
+    # @return [true,false]
+    def has_hash_attribute?(attribute)
+      hash_attributes? && existing_attributes.include?(attribute)
+    end
+
     # List of classes statically defined for this tag.
     #
     # @example For `%tag.button.button-info{ class: status }`, this returns:
@@ -197,6 +205,18 @@ module HamlLint::Tree
     # @return [String]
     def text
       (@value[:value] if @value[:parse]) || ''
+    end
+
+    private
+
+    def parsed_attributes
+      HamlLint::RubyParser.new.parse(hash_attributes_source)
+    end
+
+    def existing_attributes
+      parsed_attributes.children.collect do |parsed_attribute|
+        parsed_attribute.children.first.to_a.first
+      end
     end
   end
 end
--- a/config/default.yml
+++ b/config/default.yml
@@ -8,6 +8,9 @@
 skip_frontmatter: false
 
 linters:
+  AltText:
+    enabled: false
+
   ClassAttributeWithStaticValue:
     enabled: true
 
--- a/spec/haml_lint/linter/alt_text_spec.rb
+++ b/spec/haml_lint/linter/alt_text_spec.rb
@@ -1,0 +1,17 @@
+require 'spec_helper'
+
+describe HamlLint::Linter::AltText do
+  include_context 'linter'
+
+  context 'when there is no alt attribute' do
+    let(:haml) { '%img' }
+
+    it { should report_lint line: 1 }
+  end
+
+  context 'when there is an alt attribute' do
+    let(:haml) { '%img{ alt: "A relevant description" }' }
+
+    it { should_not report_lint }
+  end
+end
--- a/spec/haml_lint/tree/tag_node_spec.rb
+++ b/spec/haml_lint/tree/tag_node_spec.rb
@@ -5,6 +5,34 @@ describe HamlLint::Tree::TagNode do
   let(:tag_node) { parser.tree.find { |node| node.type == :tag && node.tag_name == tag_name } }
   let(:tag_name) { 'my_tag' }
 
+  describe '#has_hash_attribute?' do
+    subject { tag_node.has_hash_attribute?(:one) }
+
+    context 'when the node has the attribute' do
+      let(:haml) { '%my_tag{ one: 1 }' }
+
+      it { should == true }
+    end
+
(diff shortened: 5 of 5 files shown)


lib/haml_lint/linter/alt_text.rb
```diff
@@ -0,0 +1,11 @@
+module HamlLint
+  class Linter::AltText < Linter
+    include LinterRegistry
+
+    def visit_tag(node)
+      if node.tag_name == 'img' && !node.has_hash_attribute?(:alt)
```
sds: Are image tags the only tags that are applicable here? That seems surprising.

lib/haml_lint/tree/tag_node.rb
```diff
@@ -20,6 +20,14 @@ def contains_script?
       @value[:parse] && !@value[:value].strip.empty?
     end
 
+    # Returns whether this tag has a specified attribute
+    # `=`).
+    #
+    # @return [true,false]
+    def has_hash_attribute?(attribute)
+      hash_attributes? && eval(hash_attributes_source).has_key?(attribute)
```
sds: There is no guarantee that `eval`ing the hash attributes source will work. For example, you could have dynamic variables in the hash which of course are not defined in the context of the linter `haml-lint` run. Also, [`eval` in general is dangerous](http://stackoverflow.com/a/1903032), though that goes without saying.

You need to parse this using the `parse_ruby` helper. See https://github.com/causes/haml-lint/blob/master/lib/haml_lint/linter/class_attribute_with_static_value.rb#L30 for an example.

This is pretty new ground that I haven't developed robust abstractions for (not sure if it's really possible), so you'll need to understand and interact with the abstract syntax tree generated by the `parser` gem in order to do this properly.

sds:
Thanks for the pull request, @ckundo.

I've left some feedback inline. Main concern is around the use of `eval`, which will work only for the simplest of cases.

lib/haml_lint/linter/alt_text.rb
ckundo: This technique is specifically related to `img` tags (http://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140916/H37). There are other requirements to meet the guideline I linked to in the PR, but this is an easy test that moves towards that goal. I can make the guideline name more specific, like `ImgAltText` if that's more clear.

lib/haml_lint/tree/tag_node.rb
```diff
@@ -20,6 +20,15 @@ def contains_script?
       @value[:parse] && !@value[:value].strip.empty?
     end
 
+    # Returns whether this tag has a specified attribute
+    # `=`).
+    #
+    # @return [true,false]
+    def has_hash_attribute?(attribute)
+      hash_attributes? &&
+        hash_attributes_source.scan(/(#{attribute})\W/).any?
```
ckundo: @sds the parser gem does not give access to the hash keys as nodes, so I went with a regex. I'll refine this matcher, but is it an ok direction from your standpoint?

lib/haml_lint/linter/alt_text.rb
sds: The name is fine. I guess I was just expecting a list of different tags, so code more along the lines of:

```ruby
TAGS_REQUIRING_ALT = %w[img ...]

...

  if TAGS_REQUIRING_ALT.include?(node.tag_name) && ...
    ...
  end
```

lib/haml_lint/tree/tag_node.rb
sds: I don't think I understand what you mean by the `parser` gem not giving access to hash keys as nodes. If you look at the implementation at https://github.com/causes/haml-lint/blob/master/lib/haml_lint/linter/class_attribute_with_static_value.rb#L43 you can see we're checking to see if a key is a string or symbol.

ckundo: Yep, you're right I misunderstood. I've updated the commit.

lib/haml_lint/linter/alt_text.rb
ckundo: @sds the only tag that explicitly requires the attribute is `img`. `area` tags require them when using an `href`, and `applet` tags don't need them if they have fallback content in the tag body.

lib/haml_lint/tree/tag_node.rb
```diff
@@ -198,5 +206,17 @@ def script
     def text
       (@value[:value] if @value[:parse]) || ''
     end
+
+    private
+
+    def parsed_attributes
+      HamlLint::RubyParser.new.parse(hash_attributes_source)
```
ckundo: Seems like there can be some additional modeling/extraction done around the parser, but I wasn't sure how far to take it. Open to suggestions.

sds: There is a `parse_ruby` helper defined on the `Linter` class which you can use instead.

ckundo: @sds saw that, though it's a private method. I can promote it to public.

ckundo: also, the `parse_ruby` method is an instance method on `Linter`. Should I create a corresponding class method? It seems like adding code to the `Linter` class for the sake of `TagNode` is a smell. Should they be interacting in that way?

sds: Ah, I forgot we were in the `TagNode`. Sounds like we should define a `parse_ruby` helper on the `Node` class, since I see this being something being more heavily used in node helpers going forward.

ckundo: How about a module/mixin? That would avoid code duplication.

sds: Yup, that works.

ckundo: Hi @sds, I spiked on creating a mixin but it felt a little forced. It's less code to reuse the class here, and it's easier to test the parser object as a class vs module. 

What do you think about keeping this line as is? I think that's the only thing blocking this PR.

ckundo: Here's the spike I was talking about. Do you think this a better approach? https://github.com/ckundo/haml-lint/commit/f74eaf64b02468a3f3bef7f465504d9d0bd889bb

sds:
Merged in fe37a4c with some minor adjustments in 275fcab. Thanks!

PR #81, opened 2015-06-22
New linter: SpacesForIndentation

gburt:
Disabled by default, since tabs are acceptable in HAML, but folks may
want to enable this to enforce spaces-only.

### DIFF
--- a/spec/haml_lint/linter/spaces_for_indentation_spec.rb
+++ b/spec/haml_lint/linter/spaces_for_indentation_spec.rb
@@ -1,0 +1,45 @@
+require 'spec_helper'
+
+describe HamlLint::Linter::SpacesForIndentation do
+  include_context 'linter'
+
+  # rubocop and overcomit HardTab checks don't like tabs either, but they don't
+  # check for hex, so use that to avoid triggering them
+  let(:tab) { "\x09" }
+
+  context 'when line contains just a tab in its indentation' do
+    let(:haml) { <<-HAML }
+      %span
+#{tab * 4}- bar
+    HAML
+
+    it { should report_lint line: 2 }
+  end
+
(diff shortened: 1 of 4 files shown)


spec/haml_lint/linter/spaces_for_indentation_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+describe HamlLint::Linter::SpacesForIndentation do
+  include_context 'linter'
+
+  # rubocop and overcomit HardTab checks don't like tabs either, but they don't
+  # check for hex, so use that to avoid triggering them
+  let(:tab) { "\x09" }
```
sds: The `\t` escape sequence allows us to write tabs in Ruby without using literal tab characters, and is a little more readable.

spec/haml_lint/linter/spaces_for_indentation_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  context 'when line contains just a tab in its indentation' do
+    let(:haml) { <<-HAML }
+      %span
+#{tab * 4}- bar
+    HAML
+
+    it { should report_lint line: 2 }
+  end
+
+  context 'when line contains spaces and a tab in its indentation' do
```
sds: This test isn't necessary because the HAML parser will report an error if tabs and spaces are mixed, e.g.

```
Inconsistent indentation: "  \t" used for indentation, but the rest of the document was indented using 2 spaces.
```

sds:
Thanks for the pull request @gburt.

I've merged your change in 7707e61 and changed it to support specifying which type of indent character to enforce in 134c4146e2. The name has been changed to `Indentation` to reflect this more generic behavior.

Issue #85, opened 2015-06-25
More severity levels


skateman:
We're proudly using haml-lint in our [bot](https://github.com/ManageIQ/miq_bot) bot but we would like to use the same severity levels as [RuboCop](https://github.com/bbatsov/rubocop#severity) does. Would you accept a pull request which implements this enhancement?

sds:
Hey @skateman,

I think adding severity levels is inevitable, although I would keep it simple and only have `warning` and `error`, rather than the ones that `RuboCop` uses (`lint`, `cop`, `fatal`, etc.)

Most of the plumbing supporting this is already in place, so it should be easy to add. Feel free to submit a pull request, otherwise we'll get to it when we can.

PR #86, opened 2015-07-01
Added three new valid severities: convention, refactor and fatal

skateman:
Implements #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Option        | Description
 `enabled`     | If `false`, this linter will never be run. This takes precedence over any other option.
 `include`     | List of files or glob patterns to scope this linter to. This narrows down any files specified via the command line.
 `exclude`     | List of files or glob patterns to exclude from this linter. This excludes any files specified via the command line or already filtered via the `include` option.
-`severity`    | The severity of the linter. External tools consuming `haml-lint` output can use this to determine whether to warn or error based on the lints reported.
+`severity`    | The severity of the linter. External tools consuming `haml-lint` output can use this to determine whether to warn or error based on the lints reported. The allowed values are: `refactor`, `convention`, `warning`, `error` and `fatal`.
 
 ### Global File Exclusion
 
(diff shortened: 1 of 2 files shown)


sds:
Sorry, I've made a (silly) mistake in my response in #85. I forgot we already supported severity levels in `haml-lint` (I maintain multiple projects and got mixed up), so my response was specifically talking about adding support for severities in general. To that end, I was willing to support the `warning` and `error` severity levels, but not add additional ones.

I don't really understand the need for severity levels other than `warning` and `error`. In essence, when the linter finds a potential problem, there are one of two things you can do in response: you can decide it does not require fixing (`warning`), or you want developers to fix it (`error`). Adding additional severities doesn't make much sense to me since at some level the question still boils down to a very binary ignore/fix mentality.

Can you elaborate on why these additional levels are useful? I could definitely see value in modifying the `RuboCop` linter to support specifying how the `convention`, `refactor`, and `fatal` severities are translated into a `haml-lint` `warning`/`error` severity, but I don't see the need for `haml-lint` to support a more complex set of severities. Appreciate your thoughts on this.

Fryguy:
Hi @sds.  I can comment on the "why".  In our repo, we have a bot that runs rubocop and haml-lint together and normalizes the results into a comment in the PR.  It's a lot like houndci, but doesn't comment line-by-line.  [Here's an example](https://github.com/ManageIQ/manageiq/pull/3603#issuecomment-125285346) that does mostly haml checks.

On the rubocop side, we have a number of severity levels.  "error" generally means a Ruby syntax error or something catastrophic.  "warning" generally translates to a Ruby warning.  So, even at this point, the alignment to the haml-lint definitions of error and warning breaks down.  Then the rest fall into at least 3 other buckets that are configurable.  We've used these buckets mostly in the default to represent "must fix issues", "maybe fix issues".  For example, we've determined that "Style/ColonMethodCall: Do not use :: for method calls." is a "must fix" and we've let it at the default value of "convention".  However, something like "Line Length Too Long" isn't necessarily bad, so we've overridden the default value with "refactor" to lower its severity...this allows us to inform the author of the error without marking it as "must fix"

So, I think the underlying request is to have a way to override the severities of specific cops with custom values.  Whether those values mirror rubocop's values directly, or are from a different set, I don't think it matters, as long as we are able to configure the override.  This particular PR implements the mirroring, but I think if you wanted a different defined set like (e.g. error/warning/required/info), I think that would be fine too.

PR #92, opened 2015-09-03
Output dummy puts statements for tag closing nodes

mzp:
This fixes a bug where code like the following:

```haml
  - [1, 2, 3].each do |value|
    %div
      - if value == 0
        %div
        %div
        %div
```

...would incorrectly report a lint via the `Style/Next` cop like following:

```
foo.haml:1 [W] RuboCop: Use `next` to skip iteration.
```


eitoball:
:+1:

sds:
Thanks for the fix, @mzp!

PR #93, opened 2015-09-04
Simplify tempfile path in Rubocop linter

gylaz:
Why:

* Ruby's `Tempfile` generates unique names for files in the system's
  tmp directory.
* Providing a specific path for the temfile will fail in situations
  where the that directory doesn't exist on the filesystem.
  This is the case is virtual situations, like with Hound CI.

### DIFF
--- a/lib/haml_lint/linter/rubocop.rb
+++ b/lib/haml_lint/linter/rubocop.rb
@@ -28,10 +28,9 @@ module HamlLint
       rubocop = ::RuboCop::CLI.new
 
       original_filename = document.file || 'ruby_script'
-      filename = "#{File.basename(original_filename)}.haml_lint.tmp"
-      directory = File.dirname(original_filename)
+      filename = File.basename(original_filename)
 
-      Tempfile.open(filename, directory) do |f|
+      Tempfile.open(filename) do |f|
         begin
           f.write(ruby)
           f.close


sds:
Hey @gylaz, thanks for the pull request.

I don't think this change does what you think it does. `Tempfile.open(filename, directory)` creates a temporary file in `directory`. This is important, as when we run RuboCop against the temporary file it will automatically use the appropriate RuboCop configuration for files in that directory (since RuboCop allows you to use different `.rubocop.yml` configuration files for different directories). Moving the temporary file elsewhere prevents RuboCop from using the correct configuration on the generated Ruby file.

Thus I'm not clear on the original premise of this request&mdash;what are you seeing that suggests the directory does not exist? This should not be possible since the file is being created in the same directory as the original `.haml` file, which in theory must exist.

gylaz:
> as when we run RuboCop against the temporary file it will automatically use the appropriate RuboCop configuration for files in that directory

Ah, I hadn't considered that. Thanks for a thorough explanation!

> I don't think this change does what you think it does.

My intent was not to create the file in `directory` because it raises an exception in situations when the `directory` isn't present on the file syste. That is the case when `haml-lint` isn't ran as a cli, but its source is used for linting code (passing in the file contents). Here's an example of how we use `haml-lint` in HoundCI (we get the file content from GitHub): https://github.com/thoughtbot/hound/blob/master/app/models/style_guide/haml.rb#L40-L45

I'm not sure how best to handle it, perhaps monkey patching it for our situation would suffice. Any suggestions?

sds:
Ah, I see now. You are using `Document` in a way that wasn't intended. :)

One question: how does Hound control which RuboCop configuration to use when running `haml-lint` in this way? It looks like right now you are running it with whatever the default configuration would be.

Currently, you can set the RuboCop configuration file explicitly by [setting the `HAML_LINT_RUBOCOP_CONF` environment variable](https://github.com/brigade/haml-lint/blob/0ddaab83c4399b4c3fe13d34acd6d1c1763c8068/lib/haml_lint/linter/rubocop.rb#L74). We could add some conditional logic that generates the temporary file in the HAML file's directory only if that environment variable is not set.

Does that make sense?

gylaz:
> how does Hound control which RuboCop configuration to use when running haml-lint in this way?

The simple answer is "it doesn't". We didn't know that `haml-lint` integrated with RuboCop.
While Hound supports RuboCop for ruby files, it only supports one, top-level config.

> We could add some conditional logic that generates the temporary file in the HAML file's directory only if that environment variable is not set.

It probably would still now work in our case, because the RuboCop config is virtual (we read the content of the file from GitHub).

Looks like we'll have to find a way around this on Hound's side. Thanks for all your help!

PR #97, opened 2015-10-01
Update README. Add --reporter option.

pocke:
`haml-lint` has the `--reporter` option. 
However, this option isn't written in README.

I added it.


sds:
Thanks!

PR #106, opened 2016-01-15
Support RuboCop 0.36

maxjacobson:
RuboCop previously depended on the astrolabe gem, but in
https://github.com/bbatsov/rubocop/commit/184d1163b5662e17383735f8d008b348dd6b9d34
the dependency was removed. It seems like this was a breaking change for
haml-lint, which continues to depend on astrolabe, but doesn't
explicitly install it as part of the haml-lint installation, so
haml-lint fails when it tries to require something which isn't there.


lucascaton:
Any news on this one?

lucascaton:
@brigade Any news on this one? Thanks!

sds:
Thanks for the pull request, @maxjacobson.

Since the Astrolabe project doesn't seem to be responding to pull requests, I'm going to follow RuboCop's lead and just use their inlined implementation of S-expressions borrowed from Astrolabe.

See 5bf550f923f0bfb6f73fe77bd1a1bfa0f4f0ebf5 for more info.

maxjacobson:
@sds sounds good to me. Thanks

PR #109, opened 2016-01-20
Add explicit dependency on parser

dennisvandehoef:
We use this gem here: lib/haml_lint/ruby_parser.rb, we should also require it.

In my set up (after updating bundler and some other gems) it could not find the parser gem.
Then I found out that haml_lint uses it, but does not list it as a dependency.

### DIFF
--- a/haml_lint.gemspec
+++ b/haml_lint.gemspec
@@ -25,6 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'rake', '~> 10.0'
   s.add_dependency 'rubocop', '>= 0.36.0'
   s.add_dependency 'sysexits', '~> 1.1'
+  s.add_dependency 'parser', '>= 2.0.0'
 
   s.add_development_dependency 'rspec', '~> 3.0'
   s.add_development_dependency 'rspec-its', '~> 1.0'


sds:
Unless I'm missing something, `haml-lint`'s dependency on the `rubocop` gem should account for this. I want things to break if RuboCop stops using `parser`, since that would mean haml-lint would be using a different Ruby-parsing implementation than RuboCop.

haml-lint is pretty tightly coupled to RuboCop, so unless there's a technical reason we need to explicitly list `parser` as a dependency, I'd rather not have to remember to update the `parser` dependency every time we update haml-lint's dependency on RuboCop.

I guess I'm looking for a reason why you're seeing this behavior and I'm not? Also, can you reproduce this with the master branch of haml-lint?

dennisvandehoef:
I have done multiple updates, and the only one (from my opinion) taht could be responsible is the bundle update.

Bundler went from 1.10.6 to 1.11.2.

We updated rubocop (and its dependencies), but the latest version also has the parser gem (they updated the required version, but since that also is the version which haml_lint uses after adding it to the dependencies, that should not be the problem):

```
rubocop from 0.35.1 to 0.36.0
ast from 2.1.0 to 2.2.0
parser from 2.2.3.0 to 2.3.0.1
Removed: tins (1.6.0)
Removed:  astrolabe (1.3.1)
```
and here are the other gem updates i have done, which really should not be the problem

```
bson 2.3.0 from 3.2.6
colorize 0.7.5 to 0.7.7
connection_pool (2.1.1 to 2.2.0
ethon from 0.8.0 to 0.8.1
exception_notification 4.1.1 to 4.1.4
geocoder 1.2.12 to 1.2.14
hashie from 3.4.2 to 3.4.3
ipaddress from 0.8.0 to 0.8.2
lol_dba from 2.0.1 to 2.0.3
mini_portile from 0.6.2 to mini_portile2 2.0.0
mongoid from 4.0.0 to 4.0.2
moped from 2.0.2 to 2.0.7
newrelic_rpm from 3.9.8.273 to 3.14.1.311
nokogiri from 1.6.6.2 to 1.6.7.1
rake from 10.4.2 to 10.5.0
require_all from 1.3.2 to 1.3.3
values from 1.5.0 to 1.8.0
```

dennisvandehoef:
hmm, the error haml_lint throws when not adding the parser gem to the dependencies is:

```
/Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/gems/haml_lint-0.15.2/lib/haml_lint/ruby_parser.rb:1:in `require': cannot load such file -- astrolabe/builder (LoadError)
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/gems/haml_lint-0.15.2/lib/haml_lint/ruby_parser.rb:1:in `<top (required)>'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/gems/haml_lint-0.15.2/lib/haml_lint.rb:12:in `require'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/gems/haml_lint-0.15.2/lib/haml_lint.rb:12:in `<top (required)>'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/gems/haml_lint-0.15.2/bin/haml-lint:3:in `require'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/gems/haml_lint-0.15.2/bin/haml-lint:3:in `<top (required)>'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/bin/haml-lint:23:in `load'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/bin/haml-lint:23:in `<main>'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/bin/ruby_executable_hooks:15:in `eval'
	from /Users/dennisvandehoef/.rvm/gems/ruby-2.1.7/bin/ruby_executable_hooks:15:in `<main>'
```

dennisvandehoef:
I just tested it with the old version of bundler, got the same problem.

sds:
Hey @dennisvandehoef, the error message you posted in https://github.com/brigade/haml-lint/pull/109#issuecomment-173493877 (specifically `cannot load such file -- astrolabe/builder (LoadError)`) suggests this is a different issue, for which a fix was implemented in 5bf550f923f0bfb6f73fe77bd1a1bfa0f4f0ebf5.

I've just released `haml-lint` [0.16.0](https://github.com/brigade/haml-lint/releases/tag/v0.16.0), so give that a try and see what happens.

dennisvandehoef:
it helps, thanks.

Issue #110, opened 2016-01-22
Add Checkstyle formatter


swalberg:
Hi, Rubocop can use a checkstyle output format so that CI systems like Jenkins can pick it up, and I was wondering if there's something similar for this since it's integrated with Rubocop? Checkstyle is similar to JSON, just in XML.

The module that does it is at https://github.com/eitoball/rubocop-checkstyle_formatter but I'm not clear how the integration works.

Thanks,

Sean

sds:
Since `haml-lint` handles its own output, it can't use any custom formatters from RuboCop.

Happy to accept a pull request adding a Checkstyle formatter.

swalberg:
Thanks Shane. I'll see what I can do.

sds:
Implemented in #111. Thanks!

PR #111, opened 2016-01-23
Implement checkstyle output

swalberg:
Checkstyle output is used by various Java tools but most notably is well
supported by Jenkins CI. With the Checkstyle plugin, this change will
allow Jenkins to graph lint errors and provide deep links to the source.

jenkins 2016-01-22 19-55-07

Ref: #110

### DIFF
--- a/lib/haml_lint/reporter/checkstyle_reporter.rb
+++ b/lib/haml_lint/reporter/checkstyle_reporter.rb
@@ -1,5 +1,3 @@
-require 'builder'
-
 module HamlLint
   # Outputs report as an XML checkstyle document.
   class Reporter::CheckstyleReporter < Reporter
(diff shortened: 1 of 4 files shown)


sds:
Thanks for the pull request, @swalberg.

This from a code perspective looks great. My concern lies in the choice of third-party libraries. The [`nokogiri` gem is over 8MB](https://rubygems.org/gems/nokogiri/versions) due to its vendored libraries, so it's a pretty heavy dependency to include for such a small feature (for comparison, `haml-lint` is ~250KB). I appreciate it makes the tests easier to write, but I would prefer we avoid having to include this dependency since we're unlikely to use it for anything else.

Furthermore, the inclusion of the `builder` library is also a reasonable choice, but that project hasn't received any updates [since 2013](https://rubygems.org/gems/builder/versions/3.2.2). I'd rather not include a dependency on a library that hasn't seen _any_ development for years (there are open pull requests from 2014).

Given how simple the Checkstyle XML schema is, and given how unlikely it is (in my mind) that `haml-lint` will use an XML parsing library for anything other than this formatter, I'd prefer it if this formatter was implemented with simple string concatenation, for example what was done for the [`scss-lint` Checkstyle formatter](https://github.com/Sweetchuck/scss_lint_reporter_checkstyle/blob/31151505fb15f286b10987f6220208cb6e17a700/lib/scss_lint_reporter_checkstyle.rb). I know it's a bit cumbersome, but it's more in keeping with the needs of this particular feature.

Switch that over and I'll happily merge. Thanks!

swalberg:
Fair enough. The string concatenation is easy enough and thanks for the pointers to code.

swalberg:
Hi, very sorry about how long this took. I've made the changes and left them as a separate commit. I can squash after you've had a chance to review.

sds:
Thanks for your work on this, @swalberg!

Merged in 86a86ec.

lib/haml_lint/reporter/checkstyle_reporter.rb
```diff
@@ -0,0 +1,33 @@
...
+      log.log xml
+    end
+
+    private
+
+    def render_offenses(offenses)
+      xml = ''
+      offenses.each do |offense|
+        xml << "<error line=\"#{offense.line}\" severity=\"#{offense.severity}\" "
+        xml << "message=\"#{offense.message}\" source=\"#{offense.linter.name}\" />"
```
oliverklee: Won't this break the XML if offense.message contains a double quote?

swalberg: Almost certainly, good catch. Do you think a `gsub` will be enough to add an escape, or are there some other chars that I'm missing?

oliverklee: CGI::escapeHTML seems to be the recommended way to do so:
https://stackoverflow.com/questions/692921/how-to-html-encode-escape-a-string-is-there-a-built-in

PR #126, opened 2016-03-08
Fix middle-of-lign equals sign in UnnecessaryStringOutput

andreaswachowski:
Just stumbled upon this edge case in a project.

Note: I wondered whether changing `unnecessary_string_output.rb:25` to use `\A\s*=` is overly specific, since it excludes lines that begin with a tag, e.g. `%p= script`. But then, those lines are apparently not parsed as script nodes anyway (change `haml_visitor_spec.rb:120` to `%p= scriptB`, and that spec fails), so they would in any event never be considered by UnnecessaryStringOutput. Hope the reasoning is correct, otherwise please let me know.

### DIFF
--- a/lib/haml_lint/linter/unnecessary_string_output.rb
+++ b/lib/haml_lint/linter/unnecessary_string_output.rb
@@ -22,7 +22,7 @@ module HamlLint
     def visit_script(node)
       # Some script nodes created by the HAML parser aren't actually script
       # nodes declared via the `=` marker. Check for it.
-      return if node.source_code !~ /\s*=/
+      return if node.source_code !~ /\A\s*=/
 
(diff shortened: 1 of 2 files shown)


andreaswachowski:
@sds could you help please? The Travis build fails with a timeout during `bundle exec overcommit --run`, only for the Rubinius version though. Locally, with `rubinius 2.71828182.c1 (2.1.0 0f7189c7 2016-01-01 3.5.1 JI) [x86_64-darwin13.4.0]` (note Mac OS X instead of Linux), everything completes just fine. Thanks!

sds:
Thanks for the fix, @andreaswachowski.

You can ignore the Rubinius failure. I've merged your change in 25388e8.

Issue #124, opened 2016-02-27
Line numbers are off when using skip_frontmatter: true


agross:
I have `skip_frontmatter: true` in `.haml-lint.yml`.

Warning displayed in Atom:
> LineLength: Line is too long. [112/80] at line 9 col 5

```haml
---
title: Veranstaltungen
---
%h1
  - unless @item.fetch(:year, nil)
    Zukünftige
  Veranstaltungen
  - if @item.fetch(:year, nil)
    von                                  # Line 9 is reported as too long.
    = @item.fetch(:year)

- events = Events.new(@items).in_year(@item.fetch(:year, Date.today.year), future: @item.fetch(:year, nil).nil?)
# The last line is the actually offending line.
```

`haml-lint` also reports wrong line numbers:

```shell
$ haml-lint content/veranstaltungen.haml
content/veranstaltungen.haml:9 [W] LineLength: Line is too long. [112/80]
```

This affects editor plugins like the linters for Sublime Text and Atom.

I initially [reported the issue to Atom's linter-haml repository](https://github.com/AtomLinter/linter-haml/issues/42), but due to the fact that the shell command also shows line numbers off I think this is rather a core issue.

PR #128, opened 2016-03-15
Front matter fix

agross:
Related to #124.

My solution is to replace the frontmatter with newlines.

Another idea would be to introduce an `offset` parameter with number of frontmatter lines and pass that to the parser/nodes, but that would require much more work. I'm not aware that replacing the frontmatter with empty lines would make lints go off, despite the HAML containing unnecessary whitespace in the head.

some.haml:
``` 
---
foo: 42
--- 
%p Hello
```

With this PR merged is analyzed as:
``` 


 
%p Hello
```

### DIFF
--- a/lib/haml_lint/document.rb
+++ b/lib/haml_lint/document.rb
@@ -98,20 +98,22 @@ module HamlLint
 
     # Removes YAML frontmatter
     def strip_frontmatter(source)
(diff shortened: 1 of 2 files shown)


sds:
Thanks! 

Merged in a453f25.

PR #129, opened 2016-04-07
Checkstyle output - encoding fix

swalberg:
@oliverklee noticed that the message wasn't encoded. This fixes that problem.


swalberg:
Will re-open under a separate name.

PR #130, opened 2016-04-07
Escape the message as found by oliverklee

swalberg:
Using the stdlib CGI::escapeHTML. TIL that HTML encoding is also good
for XML! Thanks @oliverklee


sds:
Thanks for much for reporting this issue @oliverklee, and thank you @swalberg for implementing the fix!

Really appreciate the community support on this. I've gone ahead and released a hotfix in [0.17.1](https://rubygems.org/gems/haml_lint/versions/0.17.1).

PR #145, opened 2016-07-05
Configurable classes before ids

gareth:
I'm aware that `haml-lint` is an opinionated tool, I'd like to know whether there's a place in the tool for "our opinion by default, but other opinions are available".

If this isn't a direction you want to take then that's absolutely fine, but I'm also happy to work more on the implementation of what a "configurable style" looks like, if this isn't good enough for the project. For example Rubocop allows a `SupportedStyles` parameter to detect invalid style selections.

### DIFF
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -67,7 +67,9 @@ assigned.
 
 ## ClassesBeforeIds
 
-List classes before IDs in tags.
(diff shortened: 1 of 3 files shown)


sds:
Definitely encourage functionality that makes the tool more configurable. 😄 

Merged in ed0f552b1f0779a146901fd2cc00c6850a573fbe. Thanks!

PR #159, opened 2016-10-18
Disable the Metrics/BlockLength cop

oliverklee:
This RuboCop check which as introduced in RuboCop 0.44.0 does not make
sense for HAML. Hence, it is now disabled.


oliverklee:
The test failure is #161 (which is unrelated to these changes here), which I've fixed in #162.

PR #160, opened 2016-10-18
Change the Ruby version on Travis from 2.3.0 to 2.3.1

oliverklee:


oliverklee:
The test failure is #161 (which is unrelated to these changes here), which I've fixed in #162.

sds:
Thanks!

PR #162, opened 2016-10-18
Fix the Travis builds for Ruby 2.0.

### LINKED ISSUE #161: All Ruby 2.0.0 builds fail on Travis
Currently, all builds on Travis fail with Ruby 2.0.0:

https://travis-ci.org/brigade/haml-lint/jobs/168646675
https://travis-ci.org/brigade/haml-lint/jobs/168645735

Ruby 2.1, 2.2, 2.3.0 and 2.3.1 still are green.

This is because net-http-persistent requires Ruby version ~> 2.1.

net-http-persistent is a dependency of the gh gem, which is a dependency of the travis gem.

Pinning net-http-persistent might help. I'll try this.

oliverklee:
Pin net-http-persistent to version 2.9.4 as versions >= 3.0 require
Ruby >= 2.1, causing bundle install to fail.

Fixes #161

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,5 +11,11 @@ gem 'overcommit', '0.36.0'
 # Pin tool versions (which are executed by Overcommit) for Travis builds
 gem 'rubocop', '0.42.0'
 gem 'travis', '~> 1.7'
+# As long as this gem supports Ruby 2.0 (and has Travis builds for Ruby 2.0),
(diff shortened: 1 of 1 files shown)


sds:
Thanks for this fix, @oliverklee!

I think we're just going to remove the `travis` pre-commit hook, as it doesn't provide a huge amount of value, and pinning `net-http-persistent` to an old version results in our Gemnasium badge showing us as having "out of date" dependencies, which isn't strictly true.

PR #163, opened 2016-10-21
Fixed ClassesBeforeIds message when IDs first

abrom:
#### ClassesBeforeIds linter

When EnforcedStyle is IDs first, linter still reports Classes first (albeit with the failing attributes around the correct way. This change fixes that by making the message dynamic and reporting the message based on the EnforcedStyle
#### report_lint rspec matcher

To test that the above change 'worked', I also added a change to the report_lint matcher so it will fail if the lint messages don't match the expectation.

N.B. Because the matcher was getting a bit heavy, I opted to only check the message if the expected line and count were correct. This also simplified the failure message somewhat. 

I couldn't see any specs for the report_lint matcher, but I have tested a bunch of combinations of failures/messages/lines etc and it appears to work ok. Feels like this could be a change in its own right but the change to the ClassesBeforeIds change doesn't really make sense unless these matcher changes come along for the ride.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 * Fix `ClassesBeforeIds` output format to handle `id` EnforcedStyle option better
+* Allow report_lint Rspec matcher to test lint message
 
 ## 0.18.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # HAML-Lint Changelog
 
+## master (unreleased)
+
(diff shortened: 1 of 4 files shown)


abrom:
Any thoughts on this @sds  ??

sds:
Sorry for the delay in response, @abrom. I was out all last week.

Thanks for fixing this, and for adding support for message matching!

Merged in 4a82390.

sds:
Released HAML-Lint [0.18.4](https://github.com/brigade/haml-lint/releases/tag/v0.18.4)

abrom:
No problemo 👍

PR #168, opened 2016-12-13
Save temporary rubocop file as binary, closes #167

### LINKED ISSUE #167: Rubocop lints should source while preserving line-endings
I found that haml-lint does not preserve line-endings while extracting Ruby source code on Windows. In a LF file this piece of code is reported as `some.haml:1 [W] RuboCop: Style/EndOfLine: Carriage return character detected.`

some.haml:
```haml
!!!
%html{ lang: language_code_of(@item) }
```

I haven't looked at RubyExtractor, but opening the temporary file in binmode seems to fix the problem.

agross:
See #167 

Not exactly using line endings from the HAML (this would be the perfect solution). But at least we won't have to disable Style/EndOfLine to get rid of the false positives where rubocop reports CRLF offenses while the HAML source only contains LFs.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -69,6 +69,7 @@ linters:
       - Style/AlignParameters
       - Style/BlockNesting
       - Style/ElseAlignment
+      - Style/EndOfLine
       - Style/FileName
       - Style/FinalNewline
       - Style/FrozenStringLiteralComment
--- a/spec/haml_lint/linter/rubocop_spec.rb
+++ b/spec/haml_lint/linter/rubocop_spec.rb
@@ -60,4 +60,10 @@ describe HamlLint::Linter::RuboCop do
         .to have_received(:run).with(array_including('--config', 'some-rubocop.yml'))
     end
   end
+
+  context 'when running inspecting a file containing CRLF line endings (#GH-167)' do
+    let(:haml) { "- if signed_in?(viewer)\r\n%span Stuff" }
+
+    it { should_not report_lint }
+  end
 end


sds:
Thanks for digging into this issue, @agross.

I'm a little hesitant to start writing files in binary mode because that causes the [content to be treated as ASCII-8BIT](https://ruby-doc.org/core-2.3.1/IO.html#method-i-binmode). This could potentially cause problems with [HAML files containing unicode characters](https://github.com/brigade/haml-lint/issues/94).

Rather than open up a potential can of worms, since the `Style/EndOfLine` cop really doesn't make sense to keep enabled for HAML documents (HAML-Lint should check line endings itself), can you change this pull request to simply disable the cop instead?

If you think I'm missing something that your solution also addresses, please let me know. Thanks!

agross:
Hi @sds, that's all valid points. I changed the PR to just disable `Style/EndOfLine`. What's a bit sad is that I now can't really test it. Rubocop for some reason doesn't report the offense in the test, even when run on Windows. I left it though for documentation. Let me know whether that's OK with you!

lib/haml_lint/linter/rubocop.rb
```diff
@@ -71,6 +71,7 @@ def extract_lints_from_offenses(offenses, source_map)
     # @return [Array<String>]
     def rubocop_flags
       flags = %w[--format HamlLint::OffenseCollector]
+      flags += %w[--except Style/EndOfLine]
```
sds: We have a canonical list of disabled cops in the [default configuration](https://github.com/brigade/haml-lint/blob/4e7f18381f1151d9a628fcdbf632c69ca72dc72b/config/default.yml#L63-L80). The `Style/EndOfLine` cop should be added there instead. Thanks!

agross: Alright, see my updated PR.

PR #169, opened 2017-01-05
Allow templating syntax as static class

LukeeeeBennett:
There are many cases where a templating language's syntax is used in haml classes.

For example, moustache templating...

```haml
.some-class{ class: '{{ rendered_dynamic_variable }}' }
```

This currently fails the `ClassAttributeWithStaticValue` linter.

I believe these should be accounted for in the `ClassAttributeWithStaticValue` linter. The linter should ignore any class names that are not valid css syntax.

### DIFF
--- a/lib/haml_lint/linter/class_attribute_with_static_value.rb
+++ b/lib/haml_lint/linter/class_attribute_with_static_value.rb
@@ -9,10 +9,16 @@ module HamlLint
   # ...over:
   #
   #   %tag{ class: 'class-name' }
+  #
+  # But will allow invalid class names for templating:
+  #
+  #   %tag{ class: '{{ template-var }}' }
   class Linter::ClassAttributeWithStaticValue < Linter
     include LinterRegistry
 
     STATIC_TYPES = [:str, :sym].freeze
+
+    VALID_CLASS_REGEX = /^-?[_a-zA-Z]+[_a-zA-Z0-9-]*$/
 
     def visit_tag(node)
       return unless contains_class_attribute?(node.dynamic_attributes_sources)
@@ -41,7 +47,8 @@ module HamlLint
 
       STATIC_TYPES.include?(key.type) &&
         key.children.first.to_sym == :class &&
-        STATIC_TYPES.include?(value.type)
+        STATIC_TYPES.include?(value.type) &&
+        value.children.first =~ VALID_CLASS_REGEX
     end
   end
 end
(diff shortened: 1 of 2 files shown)


trotzig:
Hi @LukeeeeBennett,  and thanks for the contribution! Before I merge this however, it would be good to understand the use-case better. Could you explain your project structure, more specifically why you can't use a ruby variable in place of `'{{ rendered_dynamic_variable }}'`?

LukeeeeBennett:
Hi @trotzig! Sure, I can elaborate!

The reason we cant use a ruby variable in place of `'{{ rendered_dynamic_variable }}'` is because we actually want the string `{{ rendered_dynamic_variable }}` rendered on the frontend, just before all the JS is loaded and executed. Many frontend templating libraries will then take that rendered HTML with their special syntax (in this case, moustache brackets `{{}}`)  and they will process the HTML on the clientside to replace that special syntax with values of the same key within its own template data scope.

This goes beyond just templating libraries, as many libraries like to implement their own templating if required.

Take [Vue - Declarative Rendering](https://vuejs.org/v2/guide/#Declarative-Rendering) for example. The page is initially loaded with `{{ some_var }}` strings, which Vue, when loaded and executed, will then replace with its internal data as specified.

lib/haml_lint/linter/class_attribute_with_static_value.rb
```diff
@@ -9,11 +9,17 @@ module HamlLint
...
+  #
+  # But will allow invalid class names for templating:
+  #
+  #   %tag{ class: '{{ template-var }}' }
   class Linter::ClassAttributeWithStaticValue < Linter
     include LinterRegistry
 
     STATIC_TYPES = [:str, :sym].freeze
 
+    VALID_CLASS_REGEX = /^-?[_a-zA-Z]+[_a-zA-Z0-9-]*$/
```
trotzig: I'm a little worried that this regex won't catch all possible valid class names. That being said though, the worst thing that could happen is that you end up not seeing a lint for some obscure class name. That's probably fine.

trotzig:
Thanks @LukeeeeBennett!

lib/haml_lint/linter/class_attribute_with_static_value.rb
LukeeeeBennett: @trotzig I thought the same! The officially spec'd regex is much more complex, and it should be, but you're right the number of edge cases are significantly limited.

LukeeeeBennett:
Thanks @trotzig ❤️

PR #172, opened 2017-02-11
Switch RuboCop mode to read from stdin

michaelherold:
Currently, haml-lint runs the RuboCop linter from a Tempfile. This works
well in traditional environments, but fails in constrained environments
like Docker containers that have their data mounted as read-only
volumes.

This changes the RuboCop linter to run via an overridden STDIN. By
overriding the global `$stdin` variable, we can trick RuboCop into
believing it is reading directly from the STDIN. This means that we
don't have to generate Tempfiles for each file we want to lint which
will allow haml-lint to work within the confines of a Docker container.

This change will help me to finish my work on a [haml-lint Code Climate
engine].

[haml-lint Code Climate engine]: https://github.com/michaelherold/codeclimate-haml_lint

### DIFF
--- a/lib/haml_lint/linter/rubocop.rb
+++ b/lib/haml_lint/linter/rubocop.rb
@@ -1,6 +1,5 @@
 require 'haml_lint/ruby_extractor'
 require 'rubocop'
-require 'tempfile'
 
 module HamlLint
   # Runs RuboCop on Ruby code contained within HAML templates.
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks @michaelherold! This looks 👍 to me. Thanks for providing good context around your commit. I'm excited about the code-climate tool!

trotzig:
I prepared `0.21.0` but wasn't able to push to Rubygems. I'm going to ask @sds to add me as an owner.

trotzig:
Thanks to @sds, version `0.21.0` is now published. Thanks for the contribution!

michaelherold:
Huzzah! Thanks for the quick turnaround! :tada:

I'm going to pull in the change to my tool and make sure that it fully works. I'll keep you updated!

PR #173, opened 2017-02-14
Report syntax errors as a "Syntax" linter

michaelherold:
Currently, syntax errors are reported without a linter. This is
inconsistent with the way the rest of the lint is reported, so it makes
sense to wrap syntax errors in a similar construct as the rest of the
lint.

Since syntax errors are detected via a `Haml::Parser` that bubbles up an
exception to the `Runner`, the only place to currently put the logic is
within the `Runner` itself. This makes the `HamlLint::Linter::Syntax`
class look funny because it doesn't lint on a Haml AST - it lints on the
absence of an AST.

The benefit of this for me is that it will allow me to report syntax error
lints in my Code Climate engine, as opposed to ignoring them due to
the lack of a linter name.

### DIFF
--- a/lib/haml_lint/linter/syntax.rb
+++ b/lib/haml_lint/linter/syntax.rb
@@ -1,0 +1,6 @@
+module HamlLint
+  # A catch-all linter for syntax violations raised by the Haml parser.
+  class Linter::Syntax < Linter
+    include LinterRegistry
+  end
+end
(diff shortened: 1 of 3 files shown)


trotzig:
I'll hold off on publishing this a day or two if that's okay with you.

michaelherold:
Thanks!

Waiting is fine with me. I'll just use a Git dependency to keep moving on the Code Climate linter. This was the only thing that popped when I ran it on our production code base, so it seems like I'm close. 🎉

PR #174, opened 2017-02-15
Add support for Haml 5

### LINKED ISSUE #152: Haml gem soon to go beyond `~> 4.0`
https://github.com/brigade/haml-lint/blob/master/haml_lint.gemspec

https://www.github.com/haml/haml is currently on `5.0.0.beta.2`, will probably want to be ready to change dependency to handle 5.0 eventually.  I have had to disable this gem until then.

michaelherold:
Haml 5 is has been in Beta 2 in their repository for a while and has
some breaking API changed when compared with Haml 4. By adding an
adapter layer, we can maintain compatibility with both Haml 4 and 5.

Closes #152


lib/haml_lint/tree/tag_node.rb
```diff
@@ -186,7 +186,7 @@ def remove_inner_whitespace?
     #
     # @return [true,false]
     def remove_outer_whitespace?
-      @value[:nuke_outer_whitespace]
+      !!@value[:nuke_outer_whitespace]
```
michaelherold: This is the most controversial thing that I changed. In Haml 5, this value apparently defaults to `nil` instead of a Boolean value. This was the simplest thing that worked, so I went with it. If we want to approach it from a different angle, I'm perfectly happy to switch it up.

Appraisals
```diff
@@ -0,0 +1,7 @@
+appraise 'haml4' do
+  gem 'haml', '~> 4'
+end
+
+appraise 'haml5' do
+  gem 'haml', '~> 5.0.0.beta.2', git: 'https://github.com/haml/haml.git'
```
michaelherold: This is due to the Haml project not pushing beta tags for 5.0 to RubyGems.

michaelherold:
It looks like I have a bug on Travis that I need to fix.

lib/haml_lint/tree/tag_node.rb
trotzig: Looks fine to me.

lib/haml_lint/adapter.rb
```diff
@@ -0,0 +1,36 @@
...
+    # @api private
+    # @return [String] the approximate Haml version
+    def self.haml_version
+      Gem::Version
+        .new(Haml::VERSION)
+        .approximate_recommendation
+    end
+    private_class_method :haml_version
+  end
+end
```
trotzig: This whole class is very clean. I didn't know about `approximate_recommendation`, so I'm glad you introduced it to me. It makes the case statement easy to read and understand.

michaelherold: I was pretty excited to find it. I didn't know it existed but assumed that it did. Huzzah for `pry`ing open objects.

michaelherold:
I will definitely add information about appraisals once I figure out why Travis doesn't like its current config. Thanks for the reminder!

michaelherold:
I figured out what the problem is. Appraisal uses the `BUNDLE_GEMFILE` environment variable to pick its Gemfile from the list of candidates. The configuration for the Overcommit tool was set to override that environment variable with its own value, which prevented the `bundle install` from running.

I have disabled that part of the Overcommit configuration. I don't really like doing that, since there's a reason it was there. What do you think?

If you're okay with that change, I'll proceed with writing up a README section about Appraisals and then I'll squash this branch with that new commit.

If the change to Overcommit is too much, I'll think up another way to test against both versions!

trotzig:
This is the commit that introduced setting the `BUNDLE_GEMFILE` env variable: afe54a65acb50d6. I don't have enough context around this, so let's ping @sds, the author, to see if he can provide any good input.

trotzig:
Here's the overcommit commit that introduces that env variable: https://github.com/brigade/overcommit/commit/afe54a65acb5. I don't have a lot of context on this, so let's see if @sds (the author) can provide any helpful advise.

trotzig:
Since we haven't heard back from @sds, we just need to use our best judgement. I don't think the change you are making to .overcommit.yml will hurt, so this is good to go when we have some documentation about appraisal in the README.

michaelherold:
I added a bit about using Appraisal to the README and squashed all the commits. I also added the results of our discussion to the commit message so it will show in the `git log`.

I noticed that Travis didn't kick off on my force push. Not sure what that's about, but I thought I should mention it.

trotzig:
Thanks for the info, I'll merge and let the master build tell us if things are okay. Are you planning to do more work here or should I cut a new release?

michaelherold:
I'll probably try to knock out some more of the feature requests in the issues list. There are few low-hanging fruit there, but also a few that would require some re-architecting.

PR #175, opened 2017-02-19
Add a linter for checking id name formats

### LINKED ISSUE #138: Lint for capitalization of ids
A lint that can be configured to complain about a camelCase id would be useful.

I try to use kebab-case, but the bootstrap docs have camelCase ids in its examples.

michaelherold:
Haml style should be consistent in the naming of things. In particular,
we should be consistent in the format that our tag ids have. This linter
lets you customize the format of your ids to one of four formats:

* `lisp-case` like the Bootstrap convention (default)
* `camelCase` like the JavaScript variable convention
* `PascalCase` like the Ruby class name or React component convention
* `snake_case` like the Ruby variable convention

This is configured in the `haml-lint.yml` file as follows:

```yaml
linters:
  IdNames:
    enabled: true
    style: <STYLE>
```

where `<STYLE>` is one of: `lisp_case`, `camel_case`, `pascal_case`, or
`snake_case`.

Closes #138

### DIFF
--- a/lib/haml_lint/linter/id_names.rb
+++ b/lib/haml_lint/linter/id_names.rb
@@ -1,0 +1,28 @@
+module HamlLint
+  # Checks for `id` attributes in specific cases on tags.
+  class Linter::IdNames < Linter
+    include LinterRegistry
+
+    STYLIZED_NAMES = {
+      'camel_case' => 'camelCase',
+      'lisp_case' => 'lisp-case',
+      'pascal_case' => 'PascalCase',
+      'snake_case' => 'snake_case',
+    }.freeze
+
+    STYLES = {
+      'camel_case' => /\A[a-z][\da-zA-Z]+\z/,
+      'lisp_case' => /\A[\da-z-]+\z/,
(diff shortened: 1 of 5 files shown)


lib/haml_lint/linter/id_names.rb
```diff
@@ -0,0 +1,20 @@
...
+      'lisp_case' => /\A[\da-z-]+\z/,
+      'pascal_case' => /\A[A-Z][\da-zA-Z]+\z/,
+      'snake_case' => /\A[\da-z_]+\z/,
+    }.freeze
+
+    def visit_tag(node)
+      return unless (id = node.tag_id)
+
+      style = STYLE[config['style']] || STYLE['lisp_case']
+      record_lint(node, '`id` attribute must be in lisp-case') unless id =~ style
```
trotzig: You should replace "lisp-case" in the error message with something like `config['style'] || 'lisp-case'`

michaelherold: Good catch, thanks!

michaelherold:
I addressed that change and made the message a little more helpful by formatting the case style.

PR #176, opened 2017-02-19
Print summary of lints and files inspected

### LINKED ISSUE #123: Report total number of issues found
It would be very useful if the tool can print the number of offenses detected (like Rubocop does). So you can keep track of how many issues there are still remaining when cleaning up views.

michaelherold:
Seeing a total number of lints across all the inspected files, along
with the number of files inspected, can be helpful when you're
onboarding Haml-Lint onto a project. Since you'll likely have many
violations when starting out, seeing the total number of lints go down
over time will help motivate you to keep going.

Closes #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Command Line Flag         | Description
 `-x`/`--exclude-linter`   | Specify which linters you _don't_ want to run
 `-r`/`--reporter`         | Specify which reporter you want to use to generate the output
 `--[no-]color`            | Whether to output in color
+`--[no-]summary`          | Whether to output a summary in the default reporter
 `--show-linters`          | Show all registered linters
 `--show-reporters`        | Display available reporters
 `-h`/`--help`             | Show command line flag documentation
--- a/lib/haml_lint/cli.rb
+++ b/lib/haml_lint/cli.rb
@@ -33,7 +33,7 @@ module HamlLint
     #
     # @return [Integer] exit status code
     def act_on_options(options)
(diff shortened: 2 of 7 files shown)


michaelherold:
This actually makes it more consistent with RuboCop's output (which is what I modeled it after).

I'm curious if there are other projects that depend on the format of the default reporter. For the Code Climate engine, I used the JSON reporter, which is a lot more friendly to parsing.

trotzig:
Ah, that's good to know. On thinking about this some more, I'm actually in favor of the more verbose output, I just noticed the inconsistency with other tools. I know that overcommit [parses the output from the default reporter](https://github.com/brigade/overcommit/blob/1ecc8f5faeb8162de7ddd374ec5b1cd00c2d778a/lib/overcommit/hook/pre_commit/haml_lint.rb), and it looks like it [assumes each line of the output is either an error (default) or warning](https://github.com/brigade/overcommit/blob/1ecc8f5faeb8162de7ddd374ec5b1cd00c2d778a/lib/overcommit/hook/pre_commit/base.rb#L28). 

Maybe it would make sense to add a cli flag/config option to disable the summary?

michaelherold:
Hmm ... We could add a `-q --quiet`, but that might be too strong of a flag name. I'll think about it. 😄

michaelherold:
I went with a `--[no]-summary` option and pushed up a squashed commit with the changes.

Most of the refactoring related to `summary` was due to metrics violations from Rubocop. Apparently we ride the line a lot on `Metrics/AbcSize` and `Metrics/MethodLength`! 😄

PR #177, opened 2017-02-19
Remove extra entry in Gemfile from Haml 5 change

michaelherold:
I accidentally committed this and it got pulled in. The Haml dependency
should be specified in the appraisal, not in the base Gemfile.


trotzig:
Good catch, thanks for following up.

PR #178, opened 2017-02-20
Add --fail-level option for the CLI and Rake task

### LINKED ISSUE #137: Equivalent of --fail-level for haml-lint
RuboCop has the `--fail-level` configuration option:

> --fail-level
> 
>   Minimum severity for exit with error code. Full severity name or
> upper case initial can be given. Normally, auto-corrected offenses are
> ignored. Use A or autocorrect if you'd like them to trigger failure.

Is there an equivalent of that for haml-lint, so that even if a file has warnings, the haml-lint program returns success, so it can be used in CI?

Haml-lint has severity levels:

> severity
> 
> The severity of the linter. External tools consuming haml-lint output
> can use this to determine whether to warn or error based on the lints
> reported.

but no sign of a `--fail-level`.

(Asked on Stack Overflow at http://stackoverflow.com/questions/37535065/equivalent-of-fail-level-for-haml-lint if you want to get some internet points)

michaelherold:
This builds on the work started by @mauu-alpha in June of last year by fleshing out the functionality, extending the domain model of a severity a bit, and adding tests.

By enabling users to only fail on violations that are of the "error" severity, we can make it so you can use Haml-Lint in CI to print out style violations that are of lesser importance, but still let the build pass.

Closes #137 
Closes #142

### DIFF
--- a/lib/haml_lint/rake_task.rb
+++ b/lib/haml_lint/rake_task.rb
@@ -51,9 +51,14 @@ module HamlLint
     # @return [true,false]
     attr_accessor :quiet
 
-    # Whether output from haml-lint should not be displayed to the standard out
-    # stream.
-    # @return [true,false]
+    # The severity level above which we should fail the Rake task.
+    #
+    # @example
+    #   HamlLint::RakeTask.new do |task|
+    #     task.fail_level = 'error'
+    #   end
+    #
+    # @return [String]
     attr_accessor :fail_level
 
     # Create the task so it exists in the current namespace.
--- a/lib/haml_lint/rake_task.rb
+++ b/lib/haml_lint/rake_task.rb
@@ -51,6 +51,11 @@ module HamlLint
     # @return [true,false]
     attr_accessor :quiet
 
+    # Whether output from haml-lint should not be displayed to the standard out
+    # stream.
+    # @return [true,false]
+    attr_accessor :fail_level
+
     # Create the task so it exists in the current namespace.
     #
     # @param name [Symbol] task name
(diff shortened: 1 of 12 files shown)


lib/haml_lint/rake_task.rb
```diff
@@ -51,6 +51,11 @@ class RakeTask < Rake::TaskLib
     # @return [true,false]
     attr_accessor :quiet
 
+    # Whether output from haml-lint should not be displayed to the standard out
+    # stream.
```
trotzig: You should update this copy-pasted comment.

lib/haml_lint/severity.rb
```diff
@@ -0,0 +1,69 @@
+require 'delegate'
+
+module HamlLint
+  # Models the severity of a lint
+  class Severity < SimpleDelegator
+    include Comparable
```
trotzig: I like how you're using Comparable here so that you can use `<=` to compare different ones. Neat.

spec/haml_lint/severity_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  it 'can be matched against a symbol severity' do
+    subject.should == :warning
+  end
+
+  it 'ranks severities properly' do
+    subject.should < described_class.new(:error)
+  end
+
+  it 'can wrap itself' do
```
trotzig: I didn't fully understand why we need to support this. Can you explain?

michaelherold: This is basically a safeguard against mistakes in passing the severity level. If someone passed a `HamlLint::Severity` where we formerly accepted a symbol severity, it felt like a bad experience to throw an exception. By adding the ability for a `HamlLint::Severity` to "wrap" another, we add an easy way to ensure backward-compatibility between the old symbol-only severity indicators.

This test just ensures that the backward-compatibility doesn't break in the future.

michaelherold:
Thanks for catching that copy/paste. I cleaned it up (with an example as well!)

PR #179, opened 2017-02-28
Add a linter to catch alignment tabs

### LINKED ISSUE #153: Check for non-indentation tabs in haml
The following, which contains a tab, doesn't raise an error in haml-lint:

```
%td Hello
```

Although it probably doesn't cause any errors in HAML, I want to avoid using tabs unless there's a good reason I'm using them.

Should haml-lint have a linter for this?

michaelherold:
Even when using tabs for indentation, it is generally accepted that
using tabs for alignment is something that should be avoided. Because
you can style your tabs to look differently in your editor, there is no
guarantee that your alignment tabs will show the code as aligned in
another person's editor.

This new linter detects when tabs are being used for alignment and warns
about the presence of the tab.

Closes #153

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -9,6 +9,8 @@ PreCommit:
 
   HardTabs:
     enabled: true
+    exclude:
+      - 'lib/haml_lint/linter/README.md'
 
   RuboCop:
     enabled: true
--- a/config/default.yml
+++ b/config/default.yml
@@ -8,6 +8,9 @@
 skip_frontmatter: false
 
 linters:
+  AlignmentTabs:
+    enabled: true
+
   AltText:
     enabled: false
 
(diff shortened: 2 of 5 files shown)


.overcommit.yml
```diff
@@ -9,6 +9,8 @@ PreCommit:
 
   HardTabs:
     enabled: true
+    exclude:
+      - 'lib/haml_lint/linter/README.md'
```
michaelherold: I added this because there are examples that use hard tabs to show what not to do. I tried to figure out a way to only ignore those lines but wasn't able to find anything. If there's a way to disable overcommit's check for only specific lines, I'll be happy to make that change.

trotzig:
Thanks once again!

.overcommit.yml
trotzig: I don't think you can do that for overcommit. It makes sense to me to exclude it.

PR #180, opened 2017-03-02
Add a linter for checking instance variable usage

### LINKED ISSUE #121: No instance variables in partials
One rule of thumb that would be nice to apply is to not allow instance variables in partials. Not sure if it would make sense to have it on my default, though.

michaelherold:
It is fairly common to recommend that partials only use local variables
and helper methods and avoid the use of instance variables. This linter
checks views that follow partial conventions (prefixed with an
underscore) and checks that partials do not use instance variables.

The linter is also configurable, by default, to forbid the use of
instance variables in all views for those who wish to only use helper
methods in their views.

Lastly, it can be configured to accept any regular expression to match
against file names.

Closes #121

### DIFF
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -15,6 +15,7 @@ Below is a list of linters supported by `haml-lint`, ordered alphabetically.
 * [IdNames](#idnames)
 * [ImplicitDiv](#implicitdiv)
 * [Indentation](#indentation)
+* [InstanceVariables](#instancevariables)
 * [LeadingCommentSpace](#leadingcommentspace)
 * [LineLength](#linelength)
 * [MultilinePipe](#multilinepipe)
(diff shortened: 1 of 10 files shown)


spec/support/shared_linter_context.rb
```diff
@@ -10,9 +10,9 @@
 
   let(:config) { options[:config].for_linter(described_class) }
 
+  let(:document) { HamlLint::Document.new(normalize_indent(haml), options) }
```
michaelherold: This change lets us inject the document for the test, letting us set the file name for a test.

lib/haml_lint/tree/tag_node.rb
```diff
@@ -208,10 +222,6 @@ def script
 
     private
 
-    def parsed_attributes
```
michaelherold: Since this is a private method, I felt that I could override it with the thin `ParsedRuby` wrapper without issue.

lib/haml_lint/tree/tag_node.rb
```diff
@@ -180,6 +180,20 @@ def object_reference_source
       @value[:object_ref][/\A\[(.*)\]\z/, 1] if object_reference?
     end
 
+    # The attributes given to the tag parsed into a Ruby syntax tree.
+    #
+    # @return [ParsedRuby] syntax tree in the form returned by Parser gem
+    def parsed_attributes
+      HamlLint::ParsedRuby.new(HamlLint::RubyParser.new.parse(hash_attributes_source || ''))
```
michaelherold: Even though the original method didn't default the string to `''`, this makes it easier to use the result and shouldn't cause any side effects that I can think of.

I ran the new HEAD version against our 26,768 lines of Haml across 981 files and it didn't cause a problem.

lib/haml_lint/linter/README.md
```diff
@@ -307,6 +308,36 @@ Option          | Description
...
+You can also define your own matchers if you want to enable this linter on
+a different subset of your views. For instance, if you want to lint only files
+starting with `special_`, you can define the configuration as follows:
+
+```yaml
+InstanceVariables:
+  enabled: true
+  file_types: special
+  matchers:
+    special: ^special_.*\.haml$
```
trotzig: I like this, but I wanted to ask if you considered just using a regexp directly for `file_types`? Even though that's less readable it does avoid the added `matchers` layer. Perhaps the default `includes` configuration option would have been enough?

michaelherold: I thought about it. Mainly I went with this because I wanted a simple built-in for both the `partial` mode and the `all` mode without requiring users to write a regex for it. If you prefer the single-property architecture over this one, I'm happy to switch it over.

trotzig: Nah, I think this is fine.

PR #181, opened 2017-03-02
Add a HashReporter for other tools to use

michaelherold:
Currently, the Code Climate engine I've been working on has to go
through [a dance] to capture the JSON output of a report. By extracting
the hash construction into a reporter that can be used by tools outside
of Haml-Lint proper, we can encourage others to write such tools.

The existing structure of the JSON reporter is really nice, so this
extracts that behavior and makes the existing JSON reporter a subclass
that only handles the printing of the report to the log.

[a dance]: https://github.com/michaelherold/codeclimate-haml_lint/blob/ac22d320e95ebb476fb25551eb6fa3513d2a61b8/lib/cc/engine/report_adapter.rb


trotzig:
Thanks!

PR #182, opened 2017-03-03
Add support for incremental processing

### LINKED ISSUE #80: Add a fail-fast option
My team like to run linters as part of our builds, and consider failing linting to be a failed build. 

It would be great to add a `--fail-fast` option which would fail immediately on the first linting failure found.

### LINKED ISSUE #146: Progress Reporter?
Would love to have a progress reporter similar to what rubocop provides by default.

michaelherold:
This comes with two major features:

* The `--fail-fast` flag makes the linter fail after the first file that
  has a lint above the given fail level.
* The `--reporter progress` flag enables the use of the progress
  reporter. This new reporter prints a symbol for each linted file as it
  is linted. If the file has no lint, the reporter prints a ".", if
  there is a warning, the reporter prints a "W", and if there is an
  error the reporter prints an "E".

The incremental processing is handled via a series of hooks that are
called after certain actions within the runner. This is an extensible
system that can allow us to add more events in the future.

Closes #80
Closes #146

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ Command Line Flag         | Description
 `-i`/`--include-linter`   | Specify which linters you specifically want to run
 `-x`/`--exclude-linter`   | Specify which linters you _don't_ want to run
 `-r`/`--reporter`         | Specify which reporter you want to use to generate the output
+`--fail-fast`             | Specify whether to fail after the first file with lint
+`--fail-level`            | Specify the severity above which the lint should fail
 `--[no-]color`            | Whether to output in color
 `--[no-]summary`          | Whether to output a summary in the default reporter
 `--show-linters`          | Show all registered linters
--- a/haml_lint.gemspec
+++ b/haml_lint.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 1.9.3'
 
   s.add_dependency 'haml', '>= 4.0', '< 5.1'
+  s.add_dependency 'rainbow'
   s.add_dependency 'rake', '>= 10', '< 13'
   s.add_dependency 'rubocop', '>= 0.47.0'
   s.add_dependency 'sysexits', '~> 1.1'
--- a/lib/haml_lint/report.rb
+++ b/lib/haml_lint/report.rb
@@ -15,10 +15,28 @@ module HamlLint
     # @param lints [Array<HamlLint::Lint>] lints that were found
     # @param files [Array<String>] files that were linted
     # @param fail_level [Symbol] the severity level to fail on
-    def initialize(lints, files, fail_level = :warning)
+    # @param reporter [HamlLint::Reporter] the reporter for the report
+    def initialize(lints = [], files = [], fail_level = :warning, reporter:)
       @lints = lints.sort_by { |l| [l.filename, l.line] }
       @files = files
       @fail_level = Severity.new(fail_level)
+      @reporter = reporter
+    end
+
+    # Adds a lint to the report and notifies the reporter.
+    #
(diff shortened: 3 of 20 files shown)


README.md
```diff
@@ -71,6 +71,8 @@ Command Line Flag         | Description
 `-i`/`--include-linter`   | Specify which linters you specifically want to run
 `-x`/`--exclude-linter`   | Specify which linters you _don't_ want to run
 `-r`/`--reporter`         | Specify which reporter you want to use to generate the output
+`--fail-fast`             | Specify whether to fail after the first file with lint
+`--fail-level`            | Specify the severity above which the lint should fail
```
michaelherold: This was a piece of documentation that we missed when we added the `fail-level` option. I saw it so added it before I'd forget.

haml_lint.gemspec
```diff
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 1.9.3'
 
   s.add_dependency 'haml', '>= 4.0', '< 5.1'
+  s.add_dependency 'rainbow'
```
michaelherold: We're using the Rainbow gem, so we should probably list it as a dependency instead of relying on the transitive inclusion through Rubocop.

README.md
trotzig: 👍

haml_lint.gemspec
trotzig: Good call.

lib/haml_lint/report.rb
```diff
@@ -15,10 +15,28 @@ class Report
...
-    def initialize(lints, files, fail_level = :warning)
+    # @param reporter [HamlLint::Reporter] the reporter for the report
+    def initialize(lints = [], files = [], fail_level = :warning, reporter:)
       @lints = lints.sort_by { |l| [l.filename, l.line] }
       @files = files
       @fail_level = Severity.new(fail_level)
+      @reporter = reporter
+    end
+
+    # Add a lint to the report and notifies the reporter.
```
trotzig: Since you're using "notifies", you might want to say "Adds".

lib/haml_lint/lint.rb
```diff
@@ -1,3 +1,5 @@
+require 'haml_lint/severity'
```
trotzig: Was this added by mistake or is it something we had previously missed?

lib/haml_lint/reporter/utils.rb
```diff
@@ -0,0 +1,99 @@
...
+        log.bold lint.line, false
+      end
+
+      # Prints the severity of a lint.
+      #
+      # @param lint [HamlLint::Lint] the lint to print
+      # @return [void]
+      def print_type(lint)
+        if lint.error?
+          log.error ' [E] ', false
```
trotzig: Could we use the mark here instead? I guess that wouldn't save us much though.

michaelherold: Good call. I just pulled this directly from DefaultReporter without changes so didn't think about that.

lib/haml_lint/lint.rb
michaelherold: 🤔 I think I had worked on adding a `Severity` class the first time I tried to make the `ProgressReporter` and had it here from that. I'll check if it's necessary and remove it if it proves to not be needed.

michaelherold:
All comments addressed and squashed commit pushed!

trotzig:
Thanks! It's a pleasure to review (and merge) your PRs.

michaelherold:
It's been great to work with you. I appreciate the thorough look!

PR #183, opened 2017-03-03
Add a RepeatedId Linter

### LINKED ISSUE #120: Multiple IDs in the same template
How about a cop to check for multiple elements with the same `id` in a template? Checking for use of an id on an element in a loop might also be a nice check

michaelherold:
This linter checks for the presence of the same `id` within the view.
Since this is an error according to the HTML spec, this linter has an
`error` severity level.

Closes #120

### DIFF
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -21,6 +21,7 @@ Below is a list of linters supported by `haml-lint`, ordered alphabetically.
 * [MultilinePipe](#multilinepipe)
 * [MultilineScript](#multilinescript)
 * [ObjectReferenceAttributes](#objectreferenceattributes)
+* [RepeatedId](#repeatedid)
 * [RuboCop](#rubocop)
 * [RubyComments](#rubycomments)
 * [SpaceBeforeScript](#spacebeforescript)
@@ -477,6 +478,25 @@ where in your code a particular class attribute is defined. It is also tied
 directly to the class names of the objects you pass to it, creating an
 unnecessary coupling which can make refactoring your models affect your
 views.
+
+## RepeatedId
+
+The `id` attribute [must be unique] on the page since is intended to be a unique
+identifier. Repeating an `id` is an error in the HTML specification.
+
+**Bad**
+```haml
+#my-id
+#my-id
+```
+
+**Better**
+```haml
+#my-id
+#my-id-2
+```
+
+[must be unique]: https://www.w3.org/TR/html5/dom.html#the-id-attribute
 
 ## RuboCop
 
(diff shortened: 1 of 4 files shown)


lib/haml_lint/linter/README.md
```diff
@@ -478,6 +479,25 @@ directly to the class names of the objects you pass to it, creating an
...
+The `id` attribute [must be unique] on the page since is intended to be a unique
+identifier. Repeating an `id` is an error in the HTML specification.
+
+**Bad**
+```haml
+#my-id
+#my-id
+```
+
+**Better**
```
trotzig: Lol, I like how this isn't "Good". :)

lib/haml_lint/linter/repeated_id.rb
```diff
@@ -0,0 +1,29 @@
...
+    MESSAGE_FORMAT = %{Do not repeat id "#%s" on the page}.freeze
+
+    def visit_tag(node)
+      id = node.tag_id
+      return unless id && !id.empty?
+
+      nodes = (id_map[id] << node)
+      case nodes.size
+      when 2 then nodes.each { |logged_node| add_lint(logged_node, id) }
+      when (3..Float::INFINITY) then add_lint(node, id)
```
trotzig: It took a while for me to make sense of this, but I think I got it now. When the second id is found, add lints to all (two) lines. When the next one is found, just add a lint to that line.

michaelherold: Yeah, I went back and forth between adding the node to the list first and then checking for repeats and this version. Here's the alternative:

```ruby
def visit_tag(node)
  id = node.tag_id
  return unless id && !id.empty?

  nodes = id_map[i]
  case nodes.size
  when 0 then nodes << node
  when 1 then (nodes << node).each { |logged_node| add_lint(logged_node, id) }
  else nodes << node && add_lint(node, id)
  end
end
```

I went with the version in the PR because I wasn't a fan of the repetition of the `nodes << node`. Doing it up front means we only have to say it once.

Now that I'm thinking about it, I could do this:

```ruby
def visit_tag(node)
  id = node.tag_id
  return unless id && !id.empty?

  nodes = (id_map[id] << node)
  case nodes.size
  when 1 then return
  when 2 then nodes.each { |logged_node| add_lint(logged_node, id) }
  else add_lint(node, id)
  end
end
```

Which could then refactor to the following:

```ruby
def visit_tag(node)
  id = node.tag_id
  return unless id && !id.empty?

  nodes = (id_map[id] << node)
  case nodes.size
  when 1 then return
  when 2 then add_lint_for_first_duplication(nodes)
  else add_lint(node, id)
  end
end

private

def add_lint_for_first_duplication(nodes)
  nodes.each { |node| add_lint(node, node.tag_id) }
end
```

I think that last one is probably the winner. Do you agree?

trotzig: Yes, that would be my favorite as well.

michaelherold: Refactored, squashed and pushed!

PR #184, opened 2017-03-04
Add support for disabling linters with comments

### LINKED ISSUE #87: Add support for disable/enabling linters via comments
We need it!

michaelherold:
When you are first introducing a linting system to a code base that
hasn't not previously used one, it can be a painful experience due to
the linter raising too many problems out of the gate. By enabling
linters to be selectively toggled off within a file, we can give our
users a way to gradually introduce linting into their code base.

This feature allows you do place a Haml comment within a file to toggle
a set of linters for a block. For example, adding this comment to the
top of a Haml file:

```haml
-# haml-lint:disable all
```

will disable all of the Haml-Lint linters for the remainder of the file,
unless later enabled with a similar comment in the file, like so:

```haml
-# haml-lint:enable AltText
```

The comment line can be either the string "all" or a comma-separated
list of linter names. They will be subsequently disabled or enabled
based on the comment. In the event of conflicting directives for
a linter, the last one wins.

Closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,75 @@ rational basis, we think that the opinions themselves are less important than
 the fact that `haml-lint` provides us with an automated and low-cost means of
 enforcing consistency.
 
+## Disabling Linters within Source Code
+
+One or more individual linters can be disabled locally in a file by adding
+a directive comment. These comments look like the following:
+
+```haml
+-# haml-lint:disable AltText, LineLength
+[...]
+-# haml-lint:enable AltText, LineLength
+```
+
+You can disable *all* linters for a section with the following:
+
+```haml
+-# haml-lint:disable all
+```
+
+### Directive Scope
+
+A directive will disable the given linters for the scope of the block. This
+scope is inherited by child elements and sibling elements that come after the
+comment. For example:
+
+```haml
+-# haml-lint:disable AltText
+#content
+  %img#will-not-show-lint-1{ src: "will-not-show-lint-1.png" }
+  -# haml-lint:enable AltText
+  %img#will-show-lint-1{ src: "will-show-lint-1.png" }
+  .sidebar
+    %img#will-show-lint-2{ src: "will-show-lint-2.png" }
+%img#will-not-show-lint-2{ src: "will-not-show-lint-2.png" }
+```
+
+The `#will-not-show-lint-1` image on line 2 will not raise an `AltText` lint
+because of the directive on line 1. Since that directive is at the top level of
+the tree, it applies everywhere.
+
+However, on line 4, the directive enables the `AltText` linter for the remainder
+of the `#content` element's content. This means that the `#will-show-lint-1`
+image on line 5 will raise an `AltText` lint because it is a sibling of the
+enabling directive that appears later in the `#content` element. Likewise, the
+`#will-show-lint-2` image on line 7 will raise an `AltText` lint because it is
+a child of a sibling of the enabling directive.
+
+Lastly, the `#will-not-show-lint-2` image on line 8 will not raise an `AltText`
+lint because the enabling directive on line 4 exists in a separate element and
+is not a sibling of the it.
+
+### Directive Precedence
+
+If there are multiple directives for the same linter in an element, the last
+directive wins. For example:
+
+```haml
+-# haml-lint:enable AltText
+%p Hello, world!
+-# haml-lint:disable AltText
+%img#will-not-show-lint{ src: "will-not-show-lint.png" }
+```
+
+There are two conflicting directives for the `AltText` linter. The first one
+enables it, but the second one disables it. Since the disable directive came
+later, the `#will-not-show-lint` element will not raise an `AltText` lint.
+
+You can use this functionality to selectively enable directives within a file by
+first using the `haml-lint:disable all` directive to disable all linters in the
+file, then selectively using `haml-lint:enable` to enable linters one at a time.
+
 ## Editor Integration
 
 ### Vim
--- a/lib/haml_lint/comment_configuration.rb
+++ b/lib/haml_lint/comment_configuration.rb
@@ -1,0 +1,39 @@
+module HamlLint
+  # Determines what linters are enabled or disabled via comments.
+  class CommentConfiguration
+    # Instantiates a new {HamlLint::CommentConfiguration}.
+    #
+    # @param node [HamlLint::Tree::Node] the node to configure
+    def initialize(node)
+      @directives = node.directives.reverse
+    end
+
+    # Checks whether a linter is disabled for the node.
+    #
+    # @api public
+    # @param linter_name [String] the name of the linter
+    # @return [true, false]
+    def disabled?(linter_name)
+      most_recent_disabled = directives_for(linter_name).map(&:disable?).first
+
+      most_recent_disabled || false
+    end
+
+    private
+
+    # The list of directives in order of precedence.
+    #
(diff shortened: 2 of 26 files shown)


lib/haml_lint/linter.rb
```diff
@@ -33,7 +33,7 @@ def run(document)
     #
     # @return [String]
     def name
-      self.class.name.split('::').last
+      self.class.name.to_s.split('::').last
```
trotzig: 👍

README.md
```diff
@@ -144,6 +144,75 @@ rational basis, we think that the opinions themselves are less important than
...
+The `#will-not-show-lint-1` image on line 2 will not raise an `AltText` lint
+because of the directive on line 1. Since that directive is at the top level of
+the tree, it applies everywhere.
+
+However, on line 4, the directive enables the `AltText` linter for the remainder
+of the `#content` element's content. This means that the `#will-show-lint-1`
+image on line 5 will raise an `AltText` lint because it is a sibling of the
+enabling directive that appears later in the `#content` element. Likewise, the
+`#will-show-lint-2` image on line 7 will raise an `AltText` lint because it is
+a child of a sibling of the enabling directive.
```
trotzig: Ah, is this a common behavior amongst linters? I was sort of assuming they had no knowledge of scope. 

NINJA EDIT: looks like e.g. jshint works this way, [but not eslint](https://github.com/eslint/eslint/issues/1766).

lib/haml_lint/comment_configuration.rb
```diff
@@ -0,0 +1,40 @@
...
+    attr_reader :directives
+
+    # Finds all directives applicable to the given linter name.
+    #
+    # @api private
+    # @param linter_name [String] the name of the linter
+    # @return [Array<HamlLint::Directive>] the filtered directives
+    def directives_for(linter_name)
+      directives
+        .select { |directive| (directive.linters & ['all', linter_name]).any? }
```
trotzig: Line-breaking seems odd here. How about using a do end over multiple lines instead?

```ruby
directives.select do |directive| 
  (directive.linters & ['all', linter_name]).any?
end
```

lib/haml_lint/directive.rb
```diff
@@ -0,0 +1,123 @@
+module HamlLint
+  # Handles linter configuration transformation via Haml comments.
+  class Directive
+    LINTER_NAME_PATTERN = '(?:[A-Z]\w+)'.freeze
+    MULTIPLE_LINTERS_PATTERN = "(?:#{LINTER_NAME_PATTERN} , )*#{LINTER_NAME_PATTERN}".freeze
+    LINTERS_PATTERN = "(?<linters>all|#{MULTIPLE_LINTERS_PATTERN})".freeze
+
+    DIRECTIVE_REGEXP = Regexp.new(
+      (' haml-lint : (?<mode>(?:dis|en)able)\b ' + LINTERS_PATTERN).gsub(' ', '\s*')
```
trotzig: Minor nit: I had to read this a few times to understand what the `gsub` was achieving. I get it now (allowing any whitespace, including none). I wonder if it would have been more readable to just inline it though.

lib/haml_lint/directive.rb
```diff
@@ -0,0 +1,123 @@
...
+        Null.new(source, line)
+      end
+    end
+
+    # Instantiates a new {HamlLint::Directive}
+    #
+    # @api semipublic
+    # @param source [String] the source code to analyze
+    # @param line [Integer] the line number the source starts at
+    # @param mode [String] the type of directive
```
trotzig: Might be good to list the two valid directives here.

lib/haml_lint/tree/node.rb
```diff
@@ -95,5 +122,71 @@ def next_node
...
+      #
+      # @param node [HamlLint::Tree::Node]
+      # @return [Array<HamlLint::Tree::Node>]
+      def subsequents(node)
+        siblings[(position(node) + 1)..-1]
+      end
+
+      private
+
+      # The set of siblings with in the tree.
```
trotzig: "with in" => "within", right?

spec/haml_lint/directive_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+RSpec.describe HamlLint::Directive do
+  subject(:directive) { described_class.new }
+
+  describe '.from_line' do
+    let(:line) { 1 }
+
+    subject(:directive) { described_class.from_line(source, line) }
+
+    context 'for a line that is a directive' do
+      let(:source) { ' haml-lint:disable AltText, LineLength' }
```
trotzig: Would it make sense to test a few more strings? I'm thinking single linter, different whitespace, and perhaps one that has the linter misspelled?

spec/haml_lint/tree/haml_comment_node_spec.rb
```diff
@@ -8,7 +8,70 @@
...
+      it { should == [] }
+    end
+
+    context 'when there are directives in the node' do
+      let(:haml) { '-# haml-lint:disable AltText' }
+
+      it { should eq([HamlLint::Directive.new(haml, 1, 'disable', %w[AltText])]) }
+
+      context 'with bad formatting' do
+        let(:haml) { '-#haml-lint  :disable AltText' }
```
trotzig: Ah, ignore my previous comment. I see now that you're testing different formatting here.

README.md
michaelherold: It depends on the linter, as you pointed out.

I went with "what is the cleanest way to implement this", which was to use the AST. Since the AST is aware of scope, it automatically handles it.

Rubocop, interestingly, does not use their AST and instead tokenizes the comments separately. This seems wrong to me. Using a Haml example:

```haml
.example
  -# haml-lint:disable AltText
  %img{src: "test.png"}
%img{src: "test.png"}
```

This reads to me as "disable the AltText linter within the scope of the `.example` div". Any other behavior seems arbitrary to me.

Do you agree? My verbiage in the readme is probably pretty tortured so we can polish that, but I think the scope is a necessary part of scoping is feature because it's how I interpret the comment.

lib/haml_lint/comment_configuration.rb
michaelherold: Will do. This was leftover from when I was doing more in the block. 😄

lib/haml_lint/directive.rb
michaelherold: Yeah, I agree. The regex pattern was cribbed off of Rubocop and modified to our needs.

I can work on making it more expressive.

michaelherold:
I addressed all of the feedback except for the comment about whether the scoping was what we wanted. I figured I would wait to see what your thoughts on the current wording in the readme were.

I didn't bother to squash yet since I think there's value in the original three commits. Once we have everything hammered out I will squash the changes into the second commit.

README.md
trotzig: Sorry, I should have been more clear that I prefer it the way you have it implemented. It makes total sense.

trotzig:
Feel free to squash, or leave it as-is. Let me know and I'll merge. Also, it's starting to be time for a new release soon, don't you think?

michaelherold:
Awesome, thanks. Commits are squashed to their logical groups.

Yes, I think a release now would be good. The only issue that's still open that I have ideas about how to solve is the `--auto-gen-config` one, but I'm not sure how to address that quite yet. A release now would let us get the comment feature out in the wild, which is a major win.

PR #192, opened 2017-03-09
Update README.md to reflect ALE support for haml-lint

patricklewis:
I recently added `haml-lint` support to the [ALE plugin](https://github.com/w0rp/ale) for Vim 8 / Neovim and wanted to promote that in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,10 @@ files after saving by using the
 have the plugin, just add `let g:syntastic_haml_checkers = ['haml_lint']` to
 your `.vimrc`.
 
+### Vim 8 / Neovim
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -223,6 +223,10 @@ files after saving by using the
 have the plugin, just add `let g:syntastic_haml_checkers = ['haml_lint']` to
 your `.vimrc`.
 
+### Vim 8 / Neovim
```
michaelherold: Neomake also works out of the box with it.

I wonder at the value of these paragraphs. It seems more like the responsibility of the editor library than this readme to tell you about plugins.

trotzig: While I also question the value of these paragraphs, I think it's okay to cross-link right now. If we end up having too many of these, I might go ahead and remove them, move them to a separate page, or summarize them in some way.

PR #194, opened 2017-03-09
Ensure RepeatedId linter resets between files

### LINKED ISSUE #193: RepeatedId linter not clearing list between files
It seems that the list used by RepeatedId is not being cleared after each file is checked. I have different tabs that get activated based on anchor tags, and many different models in my Rails app have the same tabs.

Using version 0.23.0:

```
echo "#whatever Div with id attribute" > repeated_ids_1.html.haml
echo "#whatever Div with id attribute" > repeated_ids_2.html.haml
haml-lint .
```

output:

```
repeated_id_2.html.haml:1 [E] RepeatedId: Do not repeat id "#whatever" on the page
repeated_id_2.html.haml:1 [E] RepeatedId: Do not repeat id "#whatever" on the page

2 files inspected, 2 lints detected
```

michaelherold:
When I originally wrote the RepeatedId linter, I was expecting that the
linter was re-instantiated for each file. This assumption was incorrect!
That lead the linter to report repeated element IDs across different
files - a condition that normally happens in an app and does not
indicate a failure.

This fixes the issue and ensures we reset the state between different
files.

Closes #193

### DIFF
--- a/lib/haml_lint/linter/repeated_id.rb
+++ b/lib/haml_lint/linter/repeated_id.rb
@@ -5,6 +5,10 @@ module HamlLint
 
     MESSAGE_FORMAT = %{Do not repeat id "#%s" on the page}.freeze
 
+    def visit_root(_node)
+      @id_map = Hash.new { |hash, key| hash[key] = [] }
+    end
+
     def visit_tag(node)
       id = node.tag_id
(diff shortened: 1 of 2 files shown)


trotzig:
Thanks!

PR #202, opened 2017-03-18
Add support for an `inherits_from` directive

michaelherold:
Borrowing from RuboCop, the idea of inheriting a configuration from
another file is useful because it allows you to have a central
configuration for your team to use across multiple projects. It's also
a necessary feature to make an `--auto-gen-config` command useful by
giving you a configuration that you can inherit from that is generated
from an existing code base.

### DIFF
--- a/spec/haml_lint/configuration_loader_spec.rb
+++ b/spec/haml_lint/configuration_loader_spec.rb
@@ -109,6 +109,146 @@ describe HamlLint::ConfigurationLoader do
         expect { subject }.to raise_error HamlLint::Exceptions::ConfigurationError
       end
     end
+
+    context 'with an "inherits_from" directive' do
+      let(:config_file) do
+        [
+          "inherits_from: #{other_config}",
+          'skip_frontmatter: true',
+          'linters:',
+          '  AltText:',
+          '    enabled: true'
+        ].join("\n")
+      end
+      let(:other_config) { 'other_haml-lint.yml' }
+
+      before do
+        File.open(file_name, 'w') { |f| f.write(config_file) }
+      end
+
+      context 'for a file that does not exist' do
+        it 'extends the default configuration' do
+          custom_config = HamlLint::Configuration.new(
+            'inherits_from' => other_config,
+            'skip_frontmatter' => true,
+            'linters' => {
+              'AltText' => { 'enabled' => true }
+            }
+          )
+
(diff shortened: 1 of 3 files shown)


michaelherold:
🤔 I wrote this as "inherits_from" because I thought it read well, but then I noticed that RuboCop has their functionality like this as "inherit_from". I wonder if I should switch it over to match or keep the version that reads better?

spec/haml_lint/configuration_loader_spec.rb
```diff
@@ -109,6 +109,146 @@
...
+          'skip_frontmatter: true',
+          'linters:',
+          '  AltText:',
+          '    enabled: true'
+        ].join("\n")
+      end
+      let(:other_config) { 'other_haml-lint.yml' }
+
+      before do
+        File.open(file_name, 'w') { |f| f.write(config_file) }
```
trotzig: I might be missing something here, but I don't see `file_name` being defined anywhere. Also, don't you need to do some cleanup after the test is done? (bare with me, I haven't written much code in this project).

spec/haml_lint/configuration_loader_spec.rb
```diff
@@ -109,6 +109,146 @@
...
+              'AltText' => { 'enabled' => true },
+              'Indentation' => { 'enabled' => false }
+            }
+          )
+
+          subject.should ==
+            described_class.default_configuration.merge(custom_config)
+        end
+
+        context 'when the file has a circular dependency on another' do
```
trotzig: Thanks for taking care of this case!

spec/haml_lint/configuration_loader_spec.rb
trotzig: Ah, `file_name` wasn't added by you. And I take it the `around` block up top takes care of cleaning things up.

michaelherold: Yep, the `around` creates an isolated environment for creating the temp files. The tests for this class are almost to the point where they are hard to follow.

michaelherold:
Good idea on allowing both `inherit_from` and `inherits_from`. I'll add that tonight!

michaelherold:
I added `inherit_from` as an alias that combines with `inherits_from`. What do you think?

trotzig:
Looks good!

PR #203, opened 2017-03-19
Add &= and != to SpaceBeforeScript

igas:


lib/haml_lint/tree/script_node.rb
```diff
@@ -10,7 +10,7 @@ def parsed_script
       HamlLint::ParsedRuby.new(HamlLint::RubyParser.new.parse(script))
     end
 
-    # Returns the source for the script following the `-` marker.
+    # Returns the source for the script following the `=` marker.
```
michaelherold: Good catch.

lib/haml_lint/linter/space_before_script.rb
```diff
@@ -41,18 +41,23 @@ def visit_tag(node) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplex
     def visit_script(node)
       # Plain text nodes with interpolation are converted to script nodes, so we
       # need to ignore them here.
-      return unless document.source_lines[node.line - 1].lstrip.start_with?('=')
-      record_lint(node, MESSAGE_FORMAT % '=') if missing_space?(node)
+      line = document.source_lines[node.line - 1].lstrip
+      return unless line.start_with?('=', '&=', '!=')
```
michaelherold: Huh. Today I learned that you can send multiple matchers to `start_with?`. 👍

igas:
Yeah, there are few interesting differences, while in haml parser this 2 is "modifiers" to script, they behave differently, like changes in `missing_space?` for `= some_code` node text will include space, but with `&` / `!` in front of it - don't. ¯\_(ツ)_/¯

trotzig:
I agree with @michaelherold in that this might be better as a separate linter.

sds:
Closing due to lack of response. I agree this should be a separate linter for clarity's sake.

More than happy to merge once it's been extracted into a separate linter!

PR #205, opened 2017-03-20
enforce width of indentation spaces

joelbarker2011:
Ensures that HAML is indented by the proper number of spaces (default 2). If tabs are being used, then does nothing (does not infringe on the Indentation linter).

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -47,6 +47,10 @@ linters:
   ImplicitDiv:
     enabled: true
 
+  IndentationSpaces:
+    enabled: true
+    width: 2
+
   InstanceVariables:
     enabled: true
     file_types: partials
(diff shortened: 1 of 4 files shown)


trotzig:
Did you consider making this part of the `Indentation` linter? It feels a little odd to have two linters making roughly the same validations. I don't have a lot of context however, it might be more complicated merging the two than I envision it to be.

joelbarker2011:
I'm not opposed to combining them. I opted not to because the validations, while related, seemed non-overlapping, and also because I felt that the code would be more modular as a separate linter. I'll see how difficult it is to combine the two, and submit a separate PR if it works out.

trotzig:
Thanks!

joelbarker2011:
Closing this PR since #211 has been merged and and addresses the same problems better.

PR #204, opened 2017-03-19
Add `--auto-gen-config` command

michaelherold:
Onboarding a new linter onto a project can be a pain because of
preexisting violations of the style guide. To help with this, RuboCop
has their `--auto-gen-config` commands that generates a to-do list of
violations. We should have a similar command to generate the to-do list.

Closes #119
Requires #202 for the `inherits_from` command in the readme

PR #206, opened 2017-03-23
Fix bug in auto-gen-config

Ana06:
The `--auto-gen-config` option, introduced in #204 didn't work due to a bug in the `disabled_config_reporter#finished_file`. :bowtie: 

Also, when you don't specify any directory when calling `haml-lint`, it doesn't inspect any file, while I would expect it to use the current directory to look for files, as Rubocop does.

This is even more confusing when calling it with the option `auto-gen-config`, as you have to call it like:

`haml-lint . --auto-gen-config`

which is really ugly, and it is not what the documentation says. If there is any reason to not change this, I would at least add it to the documentation.

With this two changes I was able to autogenerate my `.haml-lint_todo.yml` file by running `haml-lint --auto-gen-config`.   :tada: :heart:

### DIFF
--- a/lib/haml_lint/options.rb
+++ b/lib/haml_lint/options.rb
@@ -21,7 +21,7 @@ module HamlLint
       end.parse!(args)
 
       # Any remaining arguments are assumed to be files
-      @options[:files] = args
+      @options[:files] = args.empty? ? ["."] : args
 
       @options
     rescue OptionParser::InvalidOption => ex
--- a/lib/haml_lint/reporter/disabled_config_reporter.rb
+++ b/lib/haml_lint/reporter/disabled_config_reporter.rb
@@ -66,8 +66,8 @@ module HamlLint
       if lints.any?
         lints.each do |lint|
           linters_with_lints[lint.linter.name] << lint.filename
+          linters_lint_count[lint.linter.name] += 1
         end
-        linters_lint_count[lint.linter.name] = lints.count
       end
     end
 


lib/haml_lint/options.rb
```diff
@@ -21,7 +21,7 @@ def parse(args)
       end.parse!(args)
 
       # Any remaining arguments are assumed to be files
-      @options[:files] = args
+      @options[:files] = args.empty? ? ["."] : args
```
trotzig: This changes the behavior of the CLI tool. Is it worth adding something to the README about this?

lib/haml_lint/reporter/disabled_config_reporter.rb
```diff
@@ -66,8 +66,8 @@ def finished_file(file, lints)
       if lints.any?
         lints.each do |lint|
           linters_with_lints[lint.linter.name] << lint.filename
+          linters_lint_count[lint.linter.name] = lints.count
```
michaelherold: Hah, I'm a goof.

A test to make sure this works as expected would be swell.

lib/haml_lint/options.rb
michaelherold: I'm curious, do you have a `.haml-lint.yml` file in the directory? It's odd that it doesn't work out of the box for you.

michaelherold:
I have a branch that fixes both this and the other issue you filed. I wanted to make sure you have the chance to follow up and to not steal your thunder!

If you're interested in following up on both issues, let me know and I won't submit my PR.

Thanks for taking the time to report the issues and attempt a fix! 😄

lib/haml_lint/options.rb
Ana06: I have it, but why should that make any difference? the args should be empty anyway and it will inspect just 0 files.

Ana06:
@michaelherold I had already done the fixed for the issue #208, so I'll open the PR and then we can decide what we want to merge.

regarding this PR the first commit is already included in your branch (and it was wrong here actually, I corrected). So we can merge this and then you open a PR with the rest of the changes or I can also remove the first commit and you can open a PR with your changes, I am ok with both things. We also have to decide if we want to merge the second commit.

trotzig:
I'm going to go ahead and merge this as is (including the second commit). It would be good to have a regression test for the disabled_config_reporter.rb bug though. @Ana06 - would you be willing to follow up with that?

PR #209, opened 2017-03-24
Fix duplicated entries in haml-lint_todo.yml

### LINKED ISSUE #208: duplicated entries in the autogenerated haml-lint_todo.yml
It seems that every file is added to the exclude section of the `haml-lint_todo.yml` as many times as offenses it has. For example:

```
ClassesBeforeIds:
    Exclude:
      - "app/views/webui/monitor/_events.html.haml"
      - "app/views/webui/monitor/_events.html.haml"
      - "app/views/webui/monitor/_events.html.haml"
      - "app/views/webui/monitor/_lights.html.haml"
      - "app/views/webui/monitor/_workers_table.html.haml"
      - "app/views/webui/monitor/_workers_table.html.haml"
      - "app/views/webui/package/_submit_request_dialog.html.haml"
      - "app/views/webui/package/_submit_request_dialog.html.haml"
      - "app/views/webui/package/_submit_request_dialog.html.haml"
      - "app/views/webui/patchinfo/_delete_dialog.html.haml"
      - "app/views/webui/project/subprojects.html.haml"
      - "app/views/webui/search/_tabs.html.haml"
      - "app/views/webui/theme/bento/layouts/webui/_header.html.haml"
```

**It works** but it is ugly and efficient as this will be parsed.

Ana06:
When autogenerating `haml-lint_todo.yml` It seems that every file is added to the exclude section of the haml-lint_todo.yml as many times as offenses it has.

Fixes #208

### DIFF
--- a/lib/haml_lint/reporter/disabled_config_reporter.rb
+++ b/lib/haml_lint/reporter/disabled_config_reporter.rb
@@ -65,7 +65,7 @@ module HamlLint
 
       if lints.any?
         lints.each do |lint|
-          linters_with_lints[lint.linter.name] << lint.filename
+          linters_with_lints[lint.linter.name] |= [lint.filename]
           linters_lint_count[lint.linter.name] += 1
         end
       end


Ana06:
@michaelherold you also fixed this in https://github.com/michaelherold/haml-lint/commit/9cf6c4e21f9844f0b5c4e88c3d365875a15b5442 but to be honest I prefer my solution, as including `set` is not needed.

lib/haml_lint/reporter/disabled_config_reporter.rb
```diff
@@ -65,7 +65,7 @@ def finished_file(file, lints)
 
       if lints.any?
         lints.each do |lint|
-          linters_with_lints[lint.linter.name] << lint.filename
+          linters_with_lints[lint.linter.name] |= [lint.filename]
```
trotzig: This looks wrong to me. Won't it mean that only one filename is ever added per lint type?

Ana06: No, `lints` contains every lint type as many time as offenses it finds.

michaelherold: @trotzig You're thinking of `||=` which is the "memoization method".

This is equivalent to  `linters_with_lints[lint.linter.name] = linters_with_lints[lint.linter.name] | [lint.filename]` which uses [the `#|` method](http://ruby-doc.org/core-2.4.1/Array.html#method-i-7C).

This is very elegant! I actually prefer this over using a Set. I never think to use the `#|` method, so I just reached for the heavy Set.

I can take my tests and drop them on top of this commit. What do you think @trotzig?

trotzig: Sounds good!

trotzig: Thanks for explaining the solution! Indeed, I did read it as a memoization.

PR #211, opened 2017-03-24
Indentation width

joelbarker2011:
Added support for `width` config option in the `indentation` linter. The option is only used when `character` is set to `space` (this allows the default `width` to be `2` without messing up people who are using `character: tab`, and makes more sense since probably nobody uses multiple tabs as indentation).

Includes some minor refactoring of the linter to meet Rubocop complexity standards.

Note that only the first level of indentation is checked; newer versions of `haml` now check for inconsistent indentation and raise syntax errors, and I thought it would add too much noise to flag consistently incorrect indentation throughout an entire HAML file.

This PR is a superior alternative to PR #205 and addresses the points raised therein. It also addresses another issue: #205 would have improperly allowed indentation of 4, 6, 8, etc. spaces (when `width: 2`).

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -50,6 +50,7 @@ linters:
   Indentation:
     enabled: true
     character: space # or tab
+    width: 2
 
   InstanceVariables:
     enabled: true
--- a/config/default.yml
+++ b/config/default.yml
@@ -47,6 +47,10 @@ linters:
   ImplicitDiv:
     enabled: true
 
+  Indentation:
+    enabled: true
+    character: space # or tab
+
   InstanceVariables:
     enabled: true
     file_types: partials
@@ -107,10 +111,6 @@ linters:
   SpaceInsideHashAttributes:
     enabled: true
     style: space
-
-  Indentation:
-    enabled: true
-    character: space # or tab
 
   TagName:
     enabled: true
--- a/config/default.yml
+++ b/config/default.yml
@@ -50,7 +50,7 @@ linters:
   Indentation:
     enabled: true
     character: space # or tab
-    width: 2
+    width: 2 # ignored if character == tab
 
   InstanceVariables:
     enabled: true
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -328,6 +328,8 @@ Option          | Description
   Hit me
 ```
 
+**Note:** `width` is ignored when `character` is set to `tab`.
+
 ## InstanceVariables
 
 Checks that instance variables are not used in the specified type of files.
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -308,6 +308,25 @@ Check that spaces are used for indentation instead of hard tabs.
 Option          | Description
 ----------------|-------------------------------------------------------------
 `character`     | Character to use for indentation. `space` or `tab` (default `space`)
+`width`         | Number of spaces to use for `space` indentation. (default 2)
+
+**Bad: indentation is 1 space**
+```haml
+%button
+ Hit me
+```
+
+**Bad: indentation is 4 spaces**
+```haml
+%button
+    Hit me
+```
+
+**Good: indentation is 2 spaces**
+```haml
+%button
+  Hit me
+```
 
 ## InstanceVariables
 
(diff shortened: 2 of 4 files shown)


config/default.yml
```diff
@@ -47,6 +47,11 @@ linters:
   ImplicitDiv:
     enabled: true
 
+  Indentation:
```
michaelherold: 👍  for organization!

config/default.yml
```diff
@@ -47,6 +47,11 @@ linters:
...
+    enabled: true
+    character: space # or tab
```
michaelherold: I like this lightweight documentation.

lib/haml_lint/linter/README.md
```diff
@@ -308,6 +308,25 @@ Check that spaces are used for indentation instead of hard tabs.
...
+```haml
+%button
+    Hit me
+```
+
+**Good: indentation is 2 spaces**
+```haml
+%button
+  Hit me
+```
```
michaelherold: Do you think it would be helpful to show an example of:

```yaml
Indentation:
  character: tab
  width: 1
```

I know the Ruby style guide says spaces, but there are groups that use tabs and it might help to have that example.

lib/haml_lint/linter/indentation.rb
```diff
@@ -9,15 +9,47 @@ class Linter::Indentation < Linter
...
 
+    LEADING_SPACES_REGEX = /^( +)(?! )/
+
     def visit_root(root)
-      regex = INDENT_REGEX[config['character'].to_sym]
+      character = config['character'].to_sym
+      check_character(character, root)
+
+      width = config['width'].to_i
+      check_width(width, root) if character == :space && width > 0
```
michaelherold: Oh, I see. You're only checking the width for spaces, not tabs.

I think this is probably okay since I don't see "2 tab indents" being a thing, but it might be confusing if someone tries it out.

michaelherold: I like how you're breaking the character check apart from the width check, but I noticed that you're going to be traversing the tree of every file twice because of it.

If possible, I would like to see this in one pass.

spec/haml_lint/linter/indentation_spec.rb
```diff
@@ -64,6 +113,12 @@
       HAML
 
       it { should_not report_lint }
+
+      context 'and should ignore the preferred width' do
+        let(:config) { super().merge('width' => 42) }
```
michaelherold: :joy: This would be one crazy file!

joelbarker2011:
I optimized the `width` check to return early, after finding the *first* non-disabled indented line. I also changed the message from "Line is not indented _n_ spaces" to "File does not use _n_-space indentation" so that it emphasizes that the whole *file* may be indented badly. I believe this will adequately address the speed concern.

The updated PR code is a little bizarre because of Rubocop's antipathy for breaking out of nested loops. Still, I think it's reasonably readable.

I couldn't figure out a way to use `document.source_lines` (as the `character` check uses) for checking indentation, because my code always found false positives on properly split lines like the example below.
```haml
%span { attr1: 'value1',
       attr2: 'value2' }
```
That's why I'm using `root.children` instead, and hence why I didn't combine the two checks. However, with the speed concern addressed, I think it may be more readable to have the two checks cleanly separated.

Note that `haml` already checks that the first line is not indented (`:indenting_at_start`), that indentation doesn't go deeper than one "level" at a time (`:deeper_indenting`), and that indentation is consistent (`:inconsistent_indentation`). However, `haml` does not check that the very first indentation is proper; [it assumes that it is](https://github.com/haml/haml/blob/master/lib/haml/parser.rb#L155-L156), and uses it as the benchmark for all later indentations. As long as `haml-lint` validates that the first indent is proper, it shouldn't need to check all the rest.

config/default.yml
joelbarker2011: Also added a comment here and in the README explaining that width is ignored in `tab`-mode.

lib/haml_lint/linter/README.md
joelbarker2011: Done.

lib/haml_lint/linter/indentation.rb
joelbarker2011: See long comments below.

spec/haml_lint/linter/indentation_spec.rb
trotzig: `42` is the answer to everything, so no, it would not be a crazy file. ;)

PR #218, opened 2017-04-07
Configurable consecutive comments

daveallie:
This PR add the ability to configure the maximum number of consecutive comments are allowed before the ConsectiveComment linter kicks in.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -22,6 +22,7 @@ linters:
 
   ConsecutiveComments:
     enabled: true
+    max_consecutive: 1
 
   ConsecutiveSilentScripts:
     enabled: true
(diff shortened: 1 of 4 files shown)


daveallie:
Just a preference thing, I came from 0.20, which failed to pick up on most 2 line consecutive comments in our codebase. 

Updating to the latest version of haml_lint linted all of them, but at this point, they are over our codebase. Honestly, I think two single line comments look better than a block with two lines. I'm still interested in linting more than 2 consecutive comments.

Additionally, I feel it's more inline with the ConsecutiveSilentScripts linter. (Allowing two consecutive scripts is ok, but 3 is a violation)

trotzig:
Thanks for explaining things and for the contribution!

PR #221, opened 2017-04-09
Add available options for fail level configuration

### LINKED ISSUE #216: Fail level syntax doesn't allow single letters
`bundle exec haml-lint --fail-level E` isn't regarded as valid. You have to specify "error", and the requirement that you have to use a full word isn't mentioned in the output of `bundle exec haml-lint --help`.

This is the opposite of rubocop, which allows single letters, and mentions what letters are allowed.

Possible solutions: either allow single letters, specify in the help that you have to use full words, or list what the allowed fail levels are.

agrimm:
Add the available options for fail level configuration. Also reword
documentation a little to avoid ambiguity.

Fixes #216


trotzig:
Thanks @agrimm!

Issue #224, opened 2017-04-28
undefined method `each' for nil:NilClass using Haml 5.0.0


kyohei-shimada:
haml-lint fails when I execute the following command.
Using haml 5.0.0.beta2, this problem does not occur.

```
docker-compose run web haml-lint app/views
warning: parser/current is loading parser/ruby24, which recognizes
warning: 2.4.0-compliant syntax, but you are running 2.4.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
undefined method `each' for nil:NilClass
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/ruby_extractor.rb:68:in `visit_tag'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:40:in `safe_send'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:19:in `visit'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:30:in `block in visit_children'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:30:in `each'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:30:in `visit_children'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:14:in `block in visit'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/ruby_extractor.rb:52:in `visit_root'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:40:in `safe_send'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:19:in `visit'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/ruby_extractor.rb:41:in `extract'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/linter/rubocop.rb:11:in `visit_root'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:40:in `safe_send'
/usr/local/bundle/gems/haml_lint-0.24.0/lib/haml_lint/haml_visitor.rb:19:in `visit'
[...]

PR #225, opened 2017-05-02
Fix Haml 5.0.0 compatibility

dvdbng:
Issue #224

### DIFF
--- a/lib/haml_lint/tree/tag_node.rb
+++ b/lib/haml_lint/tree/tag_node.rb
@@ -9,7 +9,7 @@ module HamlLint::Tree
     #
     # @return [Array<String>]
     def dynamic_attributes_sources
-      @value[:attributes_hashes]
+      @value.fetch(:attributes_hashes, [])
     end
 
     # Returns whether this tag contains executable script (e.g. is followed by a
--- a/Appraisals
+++ b/Appraisals
@@ -3,5 +3,5 @@ appraise 'haml4' do
 end
 
 appraise 'haml5' do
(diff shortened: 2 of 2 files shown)


trotzig:
Thanks for the contribution @dvdbng! Would you mind also updating the haml 5 version in this file? 
https://github.com/brigade/haml-lint/blob/master/Appraisals
(which I believe will make sure we catch similar things in the future).

lib/haml_lint/tree/tag_node.rb
```diff
@@ -9,7 +9,7 @@ class TagNode < Node
     #
     # @return [Array<String>]
     def dynamic_attributes_sources
-      @value[:attributes_hashes]
+      @value.fetch(:attributes_hashes, [])
```
michaelherold: 🤔  Do you know if support for this was dropped or if it was moved? I worry that we'll miss some things if we just default this to an empty array.

dvdbng: @michaelherold true, seems like it was renamed (and it's data structure changed too?) https://github.com/haml/haml/commit/7c523f278a7c62eea249ea29c797bf29377e136e
Doing this fixed haml-lint for my codebase, but likely some checks are not working properly

mensfeld:
Do you guys have any timeframe on releasing new version with that fixed? I can't use the github version unfortunately :(

sds:
@mensfeld HAML-Lint [0.25.0](https://github.com/brigade/haml-lint/releases/tag/v0.25.0) adds support for HAML 5.

mensfeld:
@sds  thank you!

PR #227, opened 2017-05-03
Add a note about usage with bundler to README.md

thutterer:


README.md
```diff
@@ -40,6 +40,12 @@ your [SCM hooks](https://github.com/brigade/overcommit).
 gem install haml_lint
 ```
 
+If you'd rather install `haml-lint` using `bundler`, don't require it in your `Gemfile`:
```
sds: This is cryptic without justification. I would explain why you do not want to `require` it (and also highlight `require` as code to emphasize we're referring to Ruby's `require` statement).

sds:
Thanks!

Issue #228, opened 2017-05-04
Support for Haml 4.1.0 beta?


phansch:
I am on `haml 4.1.0.beta.1` because of [cells-haml](https://github.com/trailblazer/cells-haml).
When I run haml-lint I get the following error:

```
$ haml-lint
Cannot handle Haml version: ~> 4.1
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/adapter.rb:21:in `detect_class'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/document.rb:48:in `process_source'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/document.rb:36:in `initialize'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:77:in `new'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:77:in `collect_lints'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:119:in `process_file'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:108:in `block in process_files'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:107:in `each'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:107:in `process_files'
/home/<snip>/.rvm/gems/ruby-2.3.3/gems/haml_lint-0.24.0/lib/haml_lint/runner.rb:132:in `report'
```
- Haml-Lint version: 0.24.0
- Haml version: 4.1.0.beta.1
- RuboCop version: 0.48.1
- Ruby version: 2.3.3

I know that haml 4.1 has been in beta for more than three years and it doesn't look like they will release a proper 4.1. It would still be nice if I could use haml-lint with this gem/version combination :sparkles:

trotzig:
[...]

PR #229, opened 2017-05-05
Add haml 4.1 to supported versions

phansch:
See #228 

Adding this allows haml-lint to start with haml 4.1.0.beta.1.
I didn't have time to check for breaking changes in that version, so it may be that it breaks something.

### DIFF
--- a/lib/haml_lint/adapter.rb
+++ b/lib/haml_lint/adapter.rb
@@ -17,6 +17,7 @@ module HamlLint
       version = haml_version
       case version
       when '~> 4.0' then HamlLint::Adapter::Haml4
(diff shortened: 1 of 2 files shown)


trotzig:
Did you figure out a way to try this out locally?

phansch:
I am currently using this branch on a private project with ~350 haml files, using haml 4.1.0.beta.1 and it  works without problems. But note that the project is not using any specific 4.1 features. 
I will also try to get the test suite running locally but I will probably not get around to that until Monday.

sds:
Merged in 02ef5c2d and added the beta to our appraisal configuration in 04906a63 so we run tests against it. Thanks for the PR!

PR #230, opened 2017-05-07
Add inline styles linter

### LINKED ISSUE #226: Suggestion: Add rule to detect inline styles
In most cases it's preferable define CSS rules in a stylesheet rather than an inline `style` attribute.

Does this seem in scope for haml-lint?

andyw8:
Closes #226

### DIFF
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -15,6 +15,7 @@ Below is a list of linters supported by `haml-lint`, ordered alphabetically.
 * [IdNames](#idnames)
 * [ImplicitDiv](#implicitdiv)
 * [Indentation](#indentation)
+* [InlineStyles](#inlinestyles)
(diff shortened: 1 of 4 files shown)


lib/haml_lint/linter/README.md
```diff
@@ -15,6 +15,7 @@ Below is a list of linters supported by `haml-lint`, ordered alphabetically.
 * [IdNames](#idnames)
 * [ImplicitDiv](#implicitdiv)
 * [Indentation](#indentation)
+* [InlineStyles](#inlinestyles)
```
andyw8: Wasn't sure if this should be `InlineStyle` or `InlineStyles`, any thoughts?

sds:
Thanks for the PR, @andyw8!

Let's keep the name as `InlineStyles` so that we can expand its responsibility to include other methods of checking inline styling, should we add any in the future.

PR #231, opened 2017-05-10
Support the stable version of Haml 5

### LINKED ISSUE #224: undefined method `each' for nil:NilClass using Haml 5.0.0

michaelherold:
Haml changed their domain model for syntax tree nodes in between
5.0.0.beta2 and 5.0.0 stable. This caused a problem with extracting the
dynamic attributes from tag nodes.

In order to maintain compatibility between Haml 4 and Haml 5, we have to
switch on the detected version of Haml and handle the dynamic attributes
differently between 4 and 5.

Haml 5 also has a subtle bug in it where the syntax tree node mutates
itself between method calls. This means that we have to memoize the
result of converting it to a literal because subsequent calls will
return a different, invalid result.

## Question

What do we think of doing the version check where I have it. I would love
to have a single place where we switch behavior based on the version of
Haml, but couldn't think of a decent way to handle it in this case. We have
[the adapter class] that we could expand, but that seems like it could get
messy quickly.

We could also add a predicate method to the adapter to do the version
check and then just keep the forks in the methods.

We could also keep doing it ad-hoc like this.

What do you all think?

Closes #224

[the adapter class]: https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/adapter.rb

### DIFF
--- a/gemfiles/haml5.gemfile
+++ b/gemfiles/haml5.gemfile
@@ -5,9 +5,9 @@ source "https://rubygems.org"
 gem "rspec", "~> 3.0"
 gem "rspec-its", "~> 1.0"
 gem "appraisal"
-gem "overcommit", "0.37.0"
-gem "rubocop", "0.47.0"
-gem "coveralls", require: false
-gem "haml", "~> 5.0.0.beta.2", git: "https://github.com/haml/haml.git"
+gem "overcommit", "0.39.1"
+gem "rubocop", "0.48.1"
+gem "coveralls", :require => false
+gem "haml", "~> 5.0.0"
 
-gemspec path: "../"
+gemspec :path => "../"
--- a/lib/haml_lint/tree/tag_node.rb
+++ b/lib/haml_lint/tree/tag_node.rb
@@ -7,9 +7,20 @@ module HamlLint::Tree
     # different attribute setting syntaxes (`{...}`/`(...)`), converted into
     # Ruby code.
     #
+    # @note This has to be memoized because of a design decision in Haml 5. When
+    # calling `DynamicAttributes#to_literal`, they mutate the "old" parameter using
+    # `String#sub!` instead of returning a new string. This means that any subsequent
+    # calls can return a nil instead of a string for that attribute, which causes
+    # any subsequent calls to the method to raise an error.
+    #
     # @return [Array<String>]
     def dynamic_attributes_sources
-      @value[:attributes_hashes]
+      @dynamic_attributes_sources ||=
+        if Gem::Version.new(Haml::VERSION) < Gem::Version.new('5')
+          @value[:attributes_hashes]
+        else
+          Array(@value[:dynamic_attributes].to_literal).reject(&:empty?)
+        end
     end
 
     # Returns whether this tag contains executable script (e.g. is followed by a
--- a/Appraisals
+++ b/Appraisals
@@ -7,5 +7,5 @@ appraise 'haml4.1-beta' do
 end
 
 appraise 'haml5' do
-  gem 'haml', '~> 5.0.0.beta.2', git: 'https://github.com/haml/haml.git'
+  gem 'haml', '~> 5.0.0'
 end
--- a/gemfiles/haml4.1_beta.gemfile
+++ b/gemfiles/haml4.1_beta.gemfile
@@ -5,9 +5,9 @@ source "https://rubygems.org"
 gem "rspec", "~> 3.0"
 gem "rspec-its", "~> 1.0"
 gem "appraisal"
-gem "overcommit", "0.37.0"
-gem "rubocop", "0.47.0"
-gem "coveralls", require: false
+gem "overcommit", "0.39.1"
+gem "rubocop", "0.48.1"
+gem "coveralls", :require => false
 gem "haml", "4.1.0.beta.1"
 
-gemspec path: "../"
+gemspec :path => "../"
--- a/gemfiles/haml4.gemfile
+++ b/gemfiles/haml4.gemfile
@@ -5,9 +5,9 @@ source "https://rubygems.org"
 gem "rspec", "~> 3.0"
 gem "rspec-its", "~> 1.0"
 gem "appraisal"
-gem "overcommit", "0.37.0"
-gem "rubocop", "0.47.0"
-gem "coveralls", require: false
+gem "overcommit", "0.39.1"
+gem "rubocop", "0.48.1"
+gem "coveralls", :require => false
 gem "haml", "~> 4"
 
-gemspec path: "../"
+gemspec :path => "../"


lib/haml_lint/tree/tag_node.rb
```diff
@@ -7,9 +7,20 @@ class TagNode < Node
...
+    #
     # @return [Array<String>]
     def dynamic_attributes_sources
-      @value[:attributes_hashes]
+      @dynamic_attributes_sources ||=
+        if Gem::Version.new(Haml::VERSION).approximate_recommendation == '~> 5.0'
+          Array(@value[:dynamic_attributes].to_literal).reject(&:empty?)
+        else
+          @value[:attributes_hashes]
+        end
```
sds: Would it make more sense to structure this to check if we are running Haml < 5 instead? The current check will fail should Haml 6 ever be released, but it's probable that the fix to get this working on Haml 5 will continue to work on Haml 6.

Designing the version check to be based on the information we have about Haml right now seems appropriate.

```ruby
@dynamic_attributes_source ||=
  if Gem::Version.new(Haml::VERSION) < Gem::Version.new('5')
    @value[:attributes_hashes]
  else
    Array(@value[:dynamic_attributes].to_literal).reject(&:empty?)
  end
```

gemfiles/haml5.gemfile
```diff
@@ -5,9 +5,9 @@ source "https://rubygems.org"
...
-gem "rubocop", "0.47.0"
-gem "coveralls", require: false
-gem "haml", "~> 5.0.0.beta.2", git: "https://github.com/haml/haml.git"
+gem "overcommit", "0.39.1"
+gem "rubocop", "0.48.1"
+gem "coveralls", :require => false
+gem "haml", "~> 5.0.0"
 
-gemspec path: "../"
+gemspec :path => "../"
```
sds: Thanks for updating these gemfiles–I had forgotten to do this.

michaelherold: Appraisal took care of this for me when I did `appraisal bundle update`. 😄

michaelherold:
Great suggestions, @sds. I amended my commit like you suggested. I think we're good to go on Haml 5 support now!

trotzig:
Thanks folks!

PR #236, opened 2017-05-19
Disable linters with many offenses in todo file

Ana06:
When autogenerating the todo file we used to list as excluded all the files with offenses. This looks horrible when having many files. :worried: 

Now when we have more than 15 files we exclude the lint instead, similarly as it is done in Rubocop. :bowtie: 

Closes https://github.com/brigade/haml-lint/issues/223

### DIFF
--- a/spec/haml_lint/reporter/disabled_config_reporter_spec.rb
+++ b/spec/haml_lint/reporter/disabled_config_reporter_spec.rb
@@ -35,26 +35,43 @@ RSpec.describe HamlLint::Reporter::DisabledConfigReporter do
     end
 
     context 'when there are lints' do
-      let(:files)        { ['some-filename.haml', 'some-filename.haml', 'other-filename.haml'] }
-      let(:lines)        { [502, 504, 724] }
-      let(:descriptions) { Array.new(3) { |i| "Description of lint #{i + 1}" } }
+      let(:files1)       { ['other-filename.haml', 'some-filename.haml', 'some-filename.haml'] }
+      let(:files2)       { [*'some-filename.haml01'..'some-filename.haml16'] }
+      let(:files)        { files1 + files2 }
+      let(:lines)        { [502, 504, 724, *1..files2.length] }
+      let(:descriptions) { Array.new(19) { |i| "Description of lint #{i + 1}" } }
       let(:header)       { output.split("\n")[0..3].join("\n") }
-      let(:linter)       { double(name: 'SomeLinter') }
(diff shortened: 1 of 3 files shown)


sds:
Hey @Ana06, thanks for the pull request!

This looks great. It would be really appreciated if we could add a test for this specific logic to ensure this behavior is preserved.

Here's an example test which should be easy to adapt for this case, though you'll likely need to create a separate `context` to capture the fact that there are more than 15 offending files: https://github.com/brigade/haml-lint/blob/1fc103a906e34baa4014cae1b5a5771d4cbe8d89/spec/haml_lint/reporter/disabled_config_reporter_spec.rb#L85-L99

Thanks again!

Ana06:
@sds 

> This looks great. It would be really appreciated if we could add a test for this specific logic to ensure this behavior is preserved.

There you have the tests! :wink: 

> Here's an example test which should be easy to adapt for this case, though you'll likely need to create a separate context to capture the fact that there are more than 15 offending files

I kept it in the text context and also add this to the other test cases (such as `prints each lint on its own line`) as it was only tested with one linter before and I think it is more precise to test it with two. I think separate it would be more complicate.

Also, I realised that the tests used to have `3 files inspected, 3 lints detected` (now `19 files inspected, 19 lints detected`). But one of the files is repeated. So I think it should be 18 and no 19 files inspected. I also added a commit for that. :wink:

spec/haml_lint/reporter/disabled_config_reporter_spec.rb
```diff
@@ -35,26 +35,43 @@
...
 
     context 'when there are lints' do
-      let(:files)        { ['some-filename.haml', 'some-filename.haml', 'other-filename.haml'] }
-      let(:lines)        { [502, 504, 724] }
-      let(:descriptions) { Array.new(3) { |i| "Description of lint #{i + 1}" } }
+      let(:files1)       { ['other-filename.haml', 'some-filename.haml', 'some-filename.haml'] }
+      let(:files2)       { [*'some-filename.haml01'..'some-filename.haml16'] }
+      let(:files)        { files1 + files2 }
+      let(:lines)        { [502, 504, 724, *1..files2.length] }
+      let(:descriptions) { Array.new(19) { |i| "Description of lint #{i + 1}" } }
```
andyw8: It seems unclear where this `19` comes from.

Ana06: `files.length` would be better, that is actually 19 here.

sds:
Thanks for adding those tests, @Ana06. Merged!

PR #237, opened 2017-05-20
Add view length linter

### LINKED ISSUE #213: Proposal: Add rule for max lines in a haml view
A view with a large number of lines is hard to manage, and should probably be split into separate partials, or have some logic extracted out, e.g. into a helpers, presenters, decorators, etc.

I'd suggest 100 lines as a starting point.

Would you accept a PR which added this?

andyw8:
Closes #213

### DIFF
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -31,6 +31,7 @@ Below is a list of linters supported by `haml-lint`, ordered alphabetically.
 * [TrailingWhitespace](#trailingwhitespace)
 * [UnnecessaryInterpolation](#unnecessaryinterpolation)
 * [UnnecessaryStringOutput](#unnecessarystringoutput)
+* [ViewLength](#viewlength)
 
 ## AlignmentTabs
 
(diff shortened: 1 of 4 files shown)


lib/haml_lint/linter/README.md
```diff
@@ -717,3 +718,13 @@ Avoid outputting string expressions in Ruby when static text will suffice.
 
 HAML gracefully handles string interpolation in static text, so you don't need
 to work with Ruby strings in order to use interpolation.
+
+## View Length
```
andyw8: It didn't seem appropriate to have good/bad examples for this linter.

lib/haml_lint/linter/view_length.rb
```diff
@@ -0,0 +1,19 @@
...
+    MSG = 'View template is too long [%d/%d]'.freeze
+
+    DummyNode = Struct.new(:line)
+
+    def visit_root(_root)
+      max = config['max']
+      line_count = document.source_lines.count
+
+      if line_count > max
+        record_lint(DummyNode.new(0), format(MSG, line_count, max))
```
andyw8: Does it make sense to report the lint on line 0?

michaelherold: 🤔 I wonder if it would be better to report on the last line of the view? It's six of one, half-dozen of the other.

michaelherold: I agree. 👍

lib/haml_lint/linter/README.md
trotzig: Agreed.

trotzig:
Thanks @andyw8!

PR #240, opened 2017-05-28
Fix UnnecessaryInterpolation for 2-char variables

malikolivier:
Two-character-long variable names would give a false positive for the
UnnecessaryInterpolation linter.

This patch fixes this issue and adds the relevant spec.

Checking the length of the script within a node is enough to do so.
As a script that contains an interpolation is at least more than 3 chars
long (#{}), limiting this linter to scripts with 3 or more char will
not have harmful effect, and will fix the issue at the same time.


trotzig:
Thanks @malikolivier for the fix and for explaining your change well.

PR #255, opened 2017-06-21
Add a note about ViewLength to v0.26.0 changelog

### LINKED ISSUE #253: ViewLength find in last release more offenses
`ViewLength` linter find offenses that were not find before when updating from `0.24.0` to `0.26.0`. It is completely fine as there is actually an offense and it should complain, but in the release notes there is nothing about this linter. Was something modified there? :confused:

michaelherold:
This adds a note (ex post facto) to the changelog for v0.26.0 about the new ViewLength linter. I updated the release notes for it but figured it should be in here too.

Closes #253


trotzig:
Thanks!

PR #258, opened 2017-08-08
Fix nested haml multiline comment unindetation

renyamizuno:
This fixes a bug where code like the following:

```haml
- some_code do
  -#
    this is comment
    multiple lines
```

would incorrectly report a lint via the `Layout/CommentIndentation` cop.

### DIFF
--- a/lib/haml_lint/ruby_extractor.rb
+++ b/lib/haml_lint/ruby_extractor.rb
@@ -173,7 +173,7 @@ module HamlLint
 
       indent = (' ' * 2 * indent_level)
 
-      @source_lines << indent + code
+      @source_lines << indentation_code(code, indent)
 
       original_line =
         node_or_line.respond_to?(:line) ? node_or_line.line : node_or_line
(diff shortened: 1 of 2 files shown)


lib/haml_lint/ruby_extractor.rb
```diff
@@ -187,6 +187,11 @@ def add_line(code, node_or_line, discard_blanks = true)
       end
     end
 
+    def indentation_code(code, indent)
```
trotzig: With a verb, this method name could be `indent_code` (which is more in line with other methods in this project).

renyamizuno: thx!
I will fix that way!

renyamizuno: Fixed it!
https://github.com/brigade/haml-lint/pull/258/commits/164cf9c8a12682e9b972c193becc1bee3a49bfbc

renyamizuno:
@trotzig thx!

PR #262, opened 2017-10-26
Update default.yml with new RuboCop names

jaredmoody:
RuboCop renamed a few cops and removed FinalNewline


jaredmoody:
Looks like the build failed because the Gemfile has Rubocop 0.49.1 pinned and these changes require 0.51.0

sds:
Thanks for the PR!

Merged in f99766b4 with some follow ups in 99e6440c...1166057c.

PR #268, opened 2018-03-30
Custom Linters

daveallie:
There is a fair amount of specific domain knowledge needed to write a linter (what HAML node you're looking for etc.) but we should at least allow people to go ahead and include their own custom linters. I previously forked this repo to add my custom linters, but I am very eager to remove that fork and just put the linters in the codebase.

This approach was very heavily influenced by Rubocop's `require` configuration.

Resolves https://github.com/brigade/haml-lint/issues/96


sds:
Thanks @daveallie!

PR #272, opened 2018-06-04
Absolute path support

daveallie:
Fix passing absolute filenames ignoring relative path excludes config. Some tools like https://github.com/okonet/lint-staged send absolute path names to the commands specified, this causes haml-lint violations when there shouldn't be any.

This PR changes `any_glob_matches?` to check both the provided filename against the ignores, along with the other of the absolute and relative path names (so absolute in the case that a relative path is passed, or relative in the case an absolute path is passed).

### Replication

Can be shown through the following test:

```yaml
# .haml-lint.yml
linters:
  InstanceVariables:
    exclude:
      - 'app/views/_my_view.html.haml'
```

```haml
-# app/views/_my_view.html.haml
= @my_ivar
```

```bash
> haml-lint app/views/_my_view.html.haml
1 file inspected, 0 lints detected

> haml-lint /path/to/repo/app/views/_my_view.html.haml
/path/to/repo/app/views/_my_view.html.haml:1 [W] InstanceVariables: Avoid using instance variables in partials views

1 file inspected, 1 lints detected
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   attributes
 * Add ability to include custom linters
 * Drop support for Ruby 2.1
+* Fix passing absolute filenames ignoring relative path excludes config
 
 ## 0.27.0
 
--- a/lib/haml_lint/utils.rb
+++ b/lib/haml_lint/utils.rb
@@ -1,3 +1,5 @@
+require 'pathname'
+
 module HamlLint
   # A miscellaneous set of utility functions.
   module Utils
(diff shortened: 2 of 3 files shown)


haml_lint.gemspec
```diff
@@ -21,9 +21,10 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.2.0'
 
+  s.add_development_dependency 'rake', '>= 10', '< 13'
+  s.add_development_dependency 'rubocop', '>= 0.50.0'
```
sds: RuboCop is not a development dependency&mdash;HAML-Lint requires it for the [RuboCop](https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/linter/rubocop.rb) linter.

The development dependency is defined in the [Gemfile](https://github.com/brigade/haml-lint/blob/a2e85796e23b25623d9cc5cffee2074ef5fc0760/Gemfile#L15).

daveallie: I actually don't know how I forgot that, I'm so used to rubocop being used internally for linting only and went on auto pilot. Will revert that change. Apologies.

daveallie: Are you happy to move rake into the Gemfile then?

daveallie: Apologies for this long PR process, have removed that commit and amended the PR body.

PR #273, opened 2018-06-20
Syntax linter should catch malformed HTML attributes

islemaster:
I'd like haml-lint's Syntax linter to catch syntax errors in HAML files, even if the RuboCop linter is disabled.

> Note: So far I've only got a failing test.  I haven't sufficiently wrapped my head around this project to spot the idiomatic way to detect malformed HAML, and would appreciate guidance with the actual fix!  Also please let me know if this test belongs in a different spec.

## Syntax error I'd like to catch
```
> echo "%tag{a: 'A',  b: 'B'}" | haml -s
<tag a='A' b='B'></tag>

> echo "%tag{a: 'A'  b: 'B'}" | haml -s
Syntax error on line 128: (haml):1: syntax error, unexpected tIDENTIFIER, expecting ')'
...ut.attributes({}, nil,a: 'A'  b: 'B')).to_s);; _hamlout.buff...
...                              ^
  Use --trace for backtrace.
```

## Context
Our project uses haml-lint but has the RuboCop linter disabled.

[Yesterday I committed and merged a HAML syntax error](https://github.com/code-dot-org/code-dot-org/pull/23195#issuecomment-398893034) on one of my projects that I expected the linter to catch.

I've got haml-lint in a precommit hook, that surfaced this linter error:

```
HtmlAttributes: Prefer the hash attributes syntax (%tag{ lang: 'en' }) over
  HTML attributes syntax (%tag(lang=en))
```

In a hasty fix, I converted
```haml
%a(href="..." target='_blank')
```
to
```haml
%a{href: "..." target='_blank'}
```
and committed the change without testing it.  This time the linter didn't complain, which I find  disappointing since it's now _completely_ broken instead of just using the wrong valid syntax.

## Changes
Moves the one Syntax linter test into its own file, and introduces a new failing test for the case I've described.

### DIFF
--- a/spec/haml_lint/linter/syntax_spec.rb
+++ b/spec/haml_lint/linter/syntax_spec.rb
@@ -1,0 +1,72 @@
+require 'spec_helper'
+
+describe HamlLint::Linter::Syntax do
+  let(:base_options) { { reporter: reporter } }
+  let(:options) { base_options }
+  let(:reporter) { HamlLint::Reporter::HashReporter.new(StringIO.new) }
+  let(:runner) { HamlLint::Runner.new }
+  let(:filename) { 'syntax_text.haml' }
+
+  include_context 'isolated environment'
+
+  before do
+    `echo "#{haml}" > #{filename}`
+    runner.stub(:extract_applicable_files).and_return([filename])
+  end
+
+  subject { runner.run(options) }
+
+  context 'when RuboCop linter is disabled' do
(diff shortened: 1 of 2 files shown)


sds:
Sorry for the lack of response. At this point in time I don't have the bandwidth to provide more assistance here (if that wasn't obvious before 😅). I am open to a working, well-tested pull request.

With that said, I'm going to close this as this hasn't had any movement—but supportive of the direction! Thanks.

PR #289, opened 2018-11-14
allow ViewLength linter be disabled inline

adrn01:
This pull request allows a user to disable the ViewLength linter on the opening line in a file. Tests are included.

### DIFF
--- a/lib/haml_lint/linter/view_length.rb
+++ b/lib/haml_lint/linter/view_length.rb
@@ -7,11 +7,12 @@ module HamlLint
 
     DummyNode = Struct.new(:line)
 
-    def visit_root(_root)
+    def visit_root(root)
       max = config['max']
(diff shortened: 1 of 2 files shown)


adrn01:
Hi @sds, thanks for your reply!
The exclude option would indeed achieve the expected result, and I can use it. However, the reason I added the inline option was because we use other inline disabling options for linters, using the [directive scope](https://github.com/brigade/haml-lint#directive-scope).
I had added the disable ViewLength to one of my files expecting it to work similarly, and was surprised to find that it hadn't. When I looked at the source I thought it had been omitted in error hence this PR.
Happy to close if you still feel it's unexpected/inconsistent behaviour.

sds:
Sorry for the delay in response, @adrn01. Fine with merging this—I had somehow missed that we had added support for "directive scopes."

PR #294, opened 2019-03-09
CI against Ruby 2.6.1

koic:
Ruby 2.6.1 has been released and this Ruby version is available on Travis CI.
https://www.ruby-lang.org/en/news/2019/01/30/ruby-2-6-1-released


sds:
Thanks @koic!

PR #296, opened 2019-04-02
Allow comment banners with LeadingCommentSpace rule

jdelStrother:
In a few of our haml templates we've ended up with banner-comments like:

```haml
-############################
-# here's an important bit! #
-############################
```

Or simply using a single line of `-#############` to break sections up a bit.

These are currently not allowed according to haml-lint's LeadingCommentSpace (although rubocop _does_ allow them).  WDYT to this change?


sds:
Seems reasonable to me—these types of comments are common. Thanks!

PR #298, opened 2019-04-08
Allow using auto-gen-config with a custom configuration file

jdelStrother:
At the moment it's not possible to use auto-gen-config with disabled or customized rules in .haml-lint.yml, since Runner just loads the default configuration.

What about this change?  By default it'll fall through to `HamlLint::ConfigurationLoader.load_applicable_config`, which will attempt to load .haml-lint.yml if it exists, and the default configuration if not.

### DIFF
--- a/lib/haml_lint/configuration_loader.rb
+++ b/lib/haml_lint/configuration_loader.rb
@@ -13,15 +13,20 @@ module HamlLint
     class << self
       # Load configuration file given the current working directory the
       # application is running within.
-      def load_applicable_config
-        directory = File.expand_path(Dir.pwd)
-        config_file = possible_config_files(directory).find(&:file?)
-
+      def load_applicable_config(config_file = nil)
+        config_file ||= default_path_to_config
         if config_file
-          load_file(config_file.to_path)
+          load_file(config_file)
         else
           default_configuration
         end
+      end
+
+      # Path to the default config file, if it exists
+      def default_path_to_config
+        directory = File.expand_path(Dir.pwd)
+        config_file = possible_config_files(directory).find(&:file?)
+        config_file ? config_file.to_path : nil
       end
 
       # Loads the built-in default configuration.
--- a/lib/haml_lint/configuration_loader.rb
+++ b/lib/haml_lint/configuration_loader.rb
@@ -13,10 +13,15 @@ module HamlLint
     class << self
       # Load configuration file given the current working directory the
       # application is running within.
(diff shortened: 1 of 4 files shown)


sds:
It would be good to add a test for this functionality if we're going to change the behavior.

Note that @michaelherold explicitly coded this behavior in #280. It would be good to chat with him about why this is preferable. I am agnostic on the matter, but since Michael did the original implementation, will defer to him.

jdelStrother:
Oh i see. I always delete the todo file before running it, so hadn’t run into that problem.
I think the easiest fix might be to have haml-lint delete the todo file automatically before starting the auto-gem run

michaelherold:
The implementation should be the same as Rubocop's. It runs with the default configuration, finds all the lints, and then should filter out the results based on your loaded configuration. Perhaps the last step isn't working correctly?

Sorry for the short message, I am on paternity leave right now so am not spending much time on open source.

jdelStrother:
OK, here's a second stab at this.  ConfigurationLoader now takes an `exclude_files` option, which is used to avoid loading the TODO file via `inherits_from`.

sds:
The tests are failing, @jdelStrother.

I'm still not clear on the value this is providing. Are we attempting to match the behavior of RuboCop, or do something different? If we're changing the behavior, let's ensure we're documenting the change in the README as well.

With that said, we should emulate RuboCop as closely as possible. Does this preserve that behavior?

jdelStrother:
> I'm still not clear on the value this is providing. Are we attempting to match the behavior of RuboCop, or do something different? If we're changing the behavior, let's ensure we're documenting the change in the README as well.

> With that said, we should emulate RuboCop as closely as possible. Does this preserve that behavior?

Sorry I wasn't clear in my initial pull request - I'm trying to fix a regression in master from v0.28.

In v0.28, we can have a .haml-lint.yml with custom rules in, and iteratively regenerate the todo file by running `rm .haml-lint_todo.yml && haml-lint --auto-gen-config`.  Removing the todo file is a workaround, otherwise zero lints are reported and your todo file ends up empty.

@michaelherold removes the need for this workaround in https://github.com/brigade/haml-lint/pull/280.  However, the way it's been fixed means that `haml-lint --auto-gen-config` no longer reads our existing .haml-lint.yml file, it just uses the default config.

This means it's not possible to use `haml-lint --auto-gen-config` with custom rules - eg changing the preferred max line length.


Does that make more sense?

Issue #302, opened 2019-05-11
require `haml_lint/spec`


LukeeeeBennett:
I really like the shared context and matchers for testing linters in the `haml-lint` repo. I would like to use them for a custom linter's spec and for now have had to copy them over to my repo.

Can these be included in the gemspec files? I am relatively new to gemfiles and ruby in general so sorry if there is already a way to do this that I have missed. If it cannot be packaged in the gem, or this is not ideal, would you accept a repo under your namespace to make spec/support reusable?

sds:
Open to a pull request adding support for this!

You'll need to move the files from `spec/support` to `lib/haml_lint/spec` and wrap them in appropriate module declarations so they don't pollute the global namespace when `require`d. 

This will ensure they are included in future published gems, since they'll be picked up by this line in the gemspec: https://github.com/sds/haml-lint/blob/024c773667e54cf88db938c2b368977005d70ee8/haml_lint.gemspec#L22

You'll then be able to require them using:
```ruby
require 'haml_lint/spec/...'
```
...or similar.

There are some other details to figure out, but happy with this approach directionally.

PR #304, opened 2019-05-13
Move spec/support to lib/haml_lint to publish

LukeeeeBennett:
With these changes, users can require useful spec/support modules to use for custom linters in their own repo.

```
require 'haml_lint/spec/normalize_indent'
require 'haml_lint/spec/shared_linter_context'
require 'haml_lint/spec/matchers/report_lint'
```

or

```
require 'haml_lint/spec'
```

Closes https://github.com/sds/haml-lint/issues/302


.rubocop.yml
```diff
@@ -65,6 +67,12 @@ Metrics/AbcSize:
 Metrics/BlockLength:
   Exclude:
     - 'spec/**/*.rb'
+    - 'lib/haml_lint/spec/**/*.rb'
+
+Metrics/ModuleLength:
```
LukeeeeBennett: Unfortunately a couple lines over because of a couple whitespace changes to make line length work.

LukeeeeBennett:
@sds This is ready for review. I made the move in a separate commit, so if you view the HEAD commit without whitespace changes it will show you every change I made to each file to get this working.

I'll note that I also wasn't sure about whether to register each component with rspec at require time in its own file (which I ended up doing) or just export a module and let importers set it up. My ruby fu is still too limited to see how to do this with the shared context or matcher so have just left it. 😂  

Ahh, maybe I dont have to worry about the context and matcher because they depend on rspec, but the normalize indent helper is a generic module and should be `include`d manually downstream?

LukeeeeBennett:
Also coveralls fails but coverage is the same. 😢 I'm guessing we can ignore that.

spec/spec_helper.rb
```diff
@@ -15,6 +15,9 @@
 Rainbow.enabled = false
 
 require 'haml_lint'
+require 'haml_lint/spec/normalize_indent'
+require 'haml_lint/spec/shared_linter_context'
+require 'haml_lint/spec/matchers/report_lint'
```
sds: You could create `lib/haml_lint/spec.rb` to contain these lines and then you would only need:

```ruby
require 'haml_lint/spec'
```

Might be helpful for other consumers as well.

I don't know how folks will use this, but I think it's a safe assumption that if you're `require`ing this that you want to have HAML Lint spec helpers injected into your RSpec runtime.

LukeeeeBennett: Awesome thanks :D

LukeeeeBennett:
@sds Back to you :)

LukeeeeBennett:
Thank you for being super responsive with guidance, review and release! 🎉 🏆

Issue #303, opened 2019-05-13
Since version 0.30, disabling LineLength doesn't work.


dmarcoux:
Hello,

First of all, thank you for `haml-lint`, it's great!

We want to update `haml-lint` to the version `0.30`, but somehow, our continuous integration is now failing due to a disable comment not working. It worked before, so this is surprising to us.

It happens in this [PR](https://github.com/openSUSE/open-build-service/pull/7527) and the continuous integration is [here](https://circleci.com/gh/openSUSE/open-build-service/24572). The offense happens in this [file](https://github.com/openSUSE/open-build-service/blob/19d4872280a149e69314d18e5ab72d2f000654a6/src/api/app/views/webui2/webui/staging/excluded_requests/index.html.haml#L7-L9).

Thank you!

Ana06:
The concrete bug is the following code, which produces an offense:

```ruby
-# haml-lint:disable LineLength
%table.responsive.table.table-sm.table-striped.table-bordered.w-100#excluded-requests-datatable{ data: excluded_requests_remote_url(@staging_workflow.id) }

sds:
Thanks for the report. Would really appreciate it if someone could take the time to perform a [bisect](https://git-scm.com/docs/git-bisect) to find the problematic commit!

rob-murray:
Hi @sds 

I'm also having this issue - it seems to be this commit: 5b4ebf93749ad091553847c18fe7239d64c62a83

I haven't had a chance to have a look at the code, just did a quick test using the files [here](https://gist.github.com/rob-murray/416bd01b2a9a1244fd39f3b4b489b6f3)

Using the commit prior to 5b4ebf93749ad091553847c18fe7239d64c62a83 works as expected

sds:
Thanks for finding the problematic commit. Tagging @michaelherold as they may have some context on what could have broken.

I'm open to reverting the commit if we can't find a resolution, but want to give folks a chance to fix first. Hoping someone can step up to help. Thanks!

michaelherold:
Thanks, @Ana06, for the reproduction. I'll take a look when I get the chance, but don't feel like you should wait for me.

michaelherold:
I pushed a commit into a pull request in #307 that takes your example line as a test case. It's not failing. I'm wondering what else is causing the problem?

Could you expand on that test case to see if we can track down a minimal reproduction script? Feel free to commit directly to that branch.

shaneog:
[...]

PR #307, opened 2019-05-23
Add a reproduction test for #303

michaelherold:
This is the reported format of line that fails on the LineLength linter.
As such, the line should fail in isolation but it doesn't fail in the
test suite. I'm going to push this up for review and solicit more
feedback from the reporters.


sds:
Fixed in #312

PR #309, opened 2019-05-28
Ignor Layout/CaseIndentation cop by default

damisul:
In Rubocop we have following settings:
Layout/CaseIndentation:
  EnforcedStyle: case
  IndentOneStep: true

In haml we have:
```
                - case item
                  - when 'home'
                    = link_to item, '#header'
```
It is a correct haml and in fact indented as required, but haml_lint produces:
RuboCop: Layout/CaseIndentation: Indent `when` one step more than `case`.

So I propose to disable this cop in default configuration

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -93,6 +93,7 @@ linters:
       - Lint/Void
       - Layout/AlignHash
       - Layout/AlignParameters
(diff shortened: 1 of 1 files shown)


sds:
Thanks!

PR #310, opened 2019-06-05
Update README with correct `auto-gen-exclude-limit` syntax and add flags to table

Zajn:
This updates usages of the `exclude-limit` option to correctly be `auto-gen-exclude-limit`, and also adds the `auto-gen`-related command line options to the table.


sds:
Thanks!

PR #312, opened 2019-08-12
Fix a bug that disabling directives are not properly applied in some cases

### LINKED ISSUE #303: Since version 0.30, disabling LineLength doesn't work.

kiyot:
This PR fixes #303 

Node#line_numbers includes line numbers of child nodes. That causes RootNode#node_for_line to return an ancestor of the node containing the specified line.

Directives are applied based on RootNode#node_for_line, so there are cases like #303 where directives are not properly applied.

#### Concerns

This PR reverts some part of #278, which is about multiline node handling.
I believe child nodes are not meant to be handled like multilines, and this PR should be okay.

### DIFF
--- a/lib/haml_lint/tree/node.rb
+++ b/lib/haml_lint/tree/node.rb
@@ -105,7 +105,7 @@ module HamlLint::Tree
       return (line..line) unless @value && text
 
       end_line = line + lines.count
-      end_line = nontrivial_end_line if line == end_line
+      end_line = nontrivial_end_line if line == end_line && children.empty?
 
       (line..end_line)
     end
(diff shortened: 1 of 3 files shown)


sds:
Thanks @kiyot. This issue seems to be affecting a decent number of folks, so appreciate a push in a direction that alleviates, though I'm not sure about this statement:
>...child nodes are not meant to be handled like multilines, and this PR should be okay.
I don't have a good sense of whether this is true/false, but I accept that an attempt at a fix is better than no action. Thanks again.

PR #313, opened 2019-09-30
Allow rake 13

jaredbeck:
AFAICT, the only breaking change in rake 13 is dropping support for ruby < 2.2.


dakotahp:
[Checking the changelog](https://github.com/ruby/rake/blob/master/History.rdoc), this is my interpretation as well. Seems safe and this gem is a blocker for me to upgrade the global rake version in my gemfile.

sds:
Thanks @jaredbeck. 

Given this is only needed for the Rake integration, I removed it entirely in 4373b517 and updated the documentation to mark it optional. I imagine most users don't utilize this integration.

Thanks!

PR #314, opened 2019-11-28
Fix problem with rubocop 0.77

kiyot:
haml-lint always fails when used with rubocop 0.77

rubocop 0.77 changed the name of Layout/TrailingBlankLines to Layout/TrailingEmptyLines.
haml-lint needs to ignore this cop, but it is not properly ignored with the latest rubocop due to the name change.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -91,13 +91,16 @@ linters:
       - Lint/BlockAlignment
       - Lint/EndAlignment
       - Lint/Void
-      - Layout/AlignHash
-      - Layout/AlignParameters
+      - Layout/AlignHash # renamed to Layout/HashAlignment in rubocop 0.77
+      - Layout/AlignParameters # renamed to Layout/ParameterAlignment in rubocop 0.77
       - Layout/CaseIndentation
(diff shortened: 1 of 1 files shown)


michaelherold:
Because Rubocop so often changes their syntax and naming, we should look into using [mry](https://github.com/pocke/mry) to make it so we don't have to chase their changes.

kiyot:
@michaelherold 

Just in case you understood, this PR is not about linting haml-lint codebase. This PR solves the problem where running haml-lint against haml files, haml-lint fails with false rubocop violations if you use rubocop 0.77.

nicoalonsop:
This PR solves https://github.com/sds/haml-lint/issues/315

michaelherold:
@kiyot Yep, I'm aware. Using `mry` means we would be able to avoid these types of changes. You can see [how CodeClimate did that in their Rubocop engine](https://github.com/codeclimate/codeclimate-rubocop/blob/master/lib/cc/engine/config_upgrader.rb) as an example for what I mean.

Issue #315, opened 2019-11-30
Layout/TrailingEmptyLines rubocop rule false positives


mockdeep:
It seems like maybe this rule should be disabled for `haml-lint`, but the Rubocop rule `Layout/TrailingEmptyLines` is reporting every one of my view files as of rubocop version 0.77.0.

mockdeep:
Looks like they renamed the rule from `Layout/TrailingBlankLines` to `Layout/TrailingEmptyLines`. Might be worth having both in `ignored_cops` for a little while.

sds:
Fixed in #314.

PR #316, opened 2019-12-04
Ignore Rubocop Layout/TrailingEmptyLines

nicoalonsop:
Solve https://github.com/sds/haml-lint/issues/315


nicoalonsop:
There is other PR that contain this.

PR #325, opened 2020-09-30
Allow haml 5.2

jaredbeck:
I'm not familiar with the reasons behind this constraint. Perhaps there is a specific reason, or perhaps it is a normal precaution. I'm just relaxing the constraint, assuming some kind of CI will run, and that'll tell us .. something.

### DIFF
--- a/haml_lint.gemspec
+++ b/haml_lint.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.4.0'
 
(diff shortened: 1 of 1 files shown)


sds:
Thanks for putting this on the radar. Needs a few more additions—addressed in #326.

jaredbeck:
Oh cool, you use Appraisal. I've been very happy with that on other projects, should've looked for the `Appraisals` file.

sds:
Released [0.36.0](https://github.com/sds/haml-lint/releases/tag/v0.36.0).

PR #328, opened 2020-11-04
Add Truffleruby head support to CI

gogainda:


.travis.yml
```diff
@@ -3,10 +3,7 @@ language: ruby
 cache: bundler
 
 rvm:
-  - 2.4
-  - 2.5
-  - 2.6
-  - 2.7
+  - truffleruby-head
```
sds, line 6: The tests fail, and I don't see why we are replacing all Ruby versions with this one.

sds:
Closing this as the tests timeout and I don't see any effort being made to fix them. Feel free to open a PR once you have everything working, but I would prefer to make this an optional build rather than required, as I personally have no interest/desire to support TruffleRuby.

gogainda:
@sds doesn't look like it's truffleruby issue, bc in the logs I dont see any truffleruby related issues. It's could be related to Travis since, but I'am not sure. Are you considering to move to Github Actions instead of Travis?

Issue #301, opened 2019-05-05
Avoid reloading .rubocop.yml repeatedly


stanhu:
I noticed that haml-lint runs `Rubocop::CLI#run` for each file, which causes Rubocop to reload all `.rubocop.yml` files repeatedly and slows down the linting process. Is there a way to reduce I/O by loading the Rubocop config file once, and then process the list of files?

sds:
Would welcome a pull request that works around this. Given this is RuboCop behavior, not sure there's an easy workaround that wouldn't tightly couple ourselves to RuboCop internals. Perhaps that's alright, but I would welcome someone else to investigate.

stanhu:
Yeah, I figured as much I've opened https://github.com/rubocop-hq/rubocop/issues/7024 to discuss. 

Another alternative would be for `haml-lint` to batch by directory, since `Rubocop::CLi#run` can handle multiple paths. What about that?

PR #322, opened 2020-04-11
Support --parallel option

sinsoku:
This commit processes files in parallel to speed them up.

### v0.35.0

```
$ time bundle exec haml-lint app/views
1158 files inspected, 0 lints detected
bundle exec haml-lint app/views  77.61s user 2.19s system 96% cpu 1:22.61 total
```

### parallel

```
$ time bundle exec haml-lint --parallel app/views
1158 files inspected, 0 lints detected
bundle exec haml-lint --parallel app/views  130.37s user 2.81s system 328% cpu 40.480 total
```

PR #329, opened 2020-12-01
Reuse a shared RuboCop::CLI instance

jdelStrother:
With ruby 2.6.6, rubocop 0.93.1, and 578 view files, this change reduced our haml-lint times from 35.9s to 17.9s

This was originally proposed in https://github.com/sds/haml-lint/issues/301.  We're still calling the entire `::RuboCop::CLI#run` method on each file, but it seems like we can get a significant speed up by re-using the same CLI instance. In particular, `@config_store` keeps a cache of config file instances, so we avoid repeatedly loading .rubocop.yml.

For each directory of files that's checked, rubocop is still traversing upwards looking for a .rubocop.yml.  You can avoid that by setting `ConfigStore#options_config`, but in practice it seemed to make very little difference, and it would prevent people from having, say, a different .rubocop.yml in `app/views/admin` vs `app/views`.

Would be curious to see if other people can reproduce the improvement.

### DIFF
--- a/lib/haml_lint/linter/rubocop.rb
+++ b/lib/haml_lint/linter/rubocop.rb
@@ -29,6 +29,14 @@ module HamlLint
 
     private
 
+    # A single CLI instance is shared between files to avoid RuboCop
+    # having to repeatedly reload .rubocop.yml.
+    def self.rubocop_cli
+      # The ivar is stored on the class singleton rather than the Linter instance
+      # because it can't be Marshal.dump'd (as used by Parallel.map)
+      @rubocop_cli ||= ::RuboCop::CLI.new
+    end
+
     # Executes RuboCop against the given Ruby code and records the offenses as
     # lints.
     #
@@ -36,8 +44,6 @@ module HamlLint
     # @param source_map [Hash] map of Ruby code line numbers to original line
     #   numbers in the template
     def find_lints(ruby, source_map)
-      @shared_rubocop ||= ::RuboCop::CLI.new
-
       filename =
         if document.file
           "#{document.file}.rb"
@@ -46,7 +52,7 @@ module HamlLint
         end
 
       with_ruby_from_stdin(ruby) do
-        extract_lints_from_offenses(lint_file(@shared_rubocop, filename), source_map)
+        extract_lints_from_offenses(lint_file(self.class.rubocop_cli, filename), source_map)
       end
     end
 
--- a/lib/haml_lint/linter/rubocop.rb
+++ b/lib/haml_lint/linter/rubocop.rb
@@ -36,7 +36,7 @@ module HamlLint
     # @param source_map [Hash] map of Ruby code line numbers to original line
     #   numbers in the template
     def find_lints(ruby, source_map)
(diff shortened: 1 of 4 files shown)


jdelStrother:
Whoops, this doesn't play nicely with `--parallel`. Fix incoming...

sds:
This still in flight, @jdelStrother? Supportive of the change, but will close if unable to address the broken test and the `--parallel` support.

jdelStrother:
Oh sorry - I was seeing broken tests on master & assumed the errors here were the same thing.
I've pushed an [amended commit](https://github.com/sds/haml-lint/pull/329/commits/13e18e91a79d9883442967ec46670b4b7b9032cf) that should fix the errors I introduced, plus an [additional commit](https://github.com/sds/haml-lint/pull/329/commits/f376704c0b8ecfe53f51ed042260f8f2c12eb4d4) that fixes an existing broken test under HAML 5.2

I'm also seeing this error in master (and on my branch):
```
  1) HamlLint::Linter::InstanceVariables when the partial is actually an ERB file that writes Haml is expected to report a lint on line 1 with message 'unterminated string meets end of file'
     Failure/Error: it { should report_lint line: 1, message: 'unterminated string meets end of file' }
       expected that a lint would be reported on line 1 with message 'unterminated string meets end of file'
     # ./spec/haml_lint/linter/instance_variables_spec.rb:113:in `block (3 levels) in <top (required)>'

Finished in 1.56 seconds (files took 1.44 seconds to load)
683 examples, 1 failure

Failed examples:

rspec ./spec/haml_lint/linter/instance_variables_spec.rb:113 # HamlLint::Linter::InstanceVariables when the partial is actually an ERB file that writes Haml is expected to report a lint on line 1 with message 'unterminated string meets end of file'

```

which I'm struggling to pin down. I _think_ it's caused by differences in the `parser` gem between 2.7 & 3.0, but confused why it doesn't seem to produce the same error on Travis.

sds:
Thanks for digging in. I don't have any obvious suggestions based on what we're seeing, and at the moment don't have bandwidth to dig into this deeper. If you are able to figure this out, I will happily merge.

jdelStrother:
So, the error is also present on master. The test suite used to pass back in October when it was last run, but since then v3 of the Parser gem has come out, so new test runs fail due to that, when run in ruby 2.7.

The new behaviour actually seems like an improvement - with Ruby 2.7 & Parser v3, it can give sensible HAML linting errors on erb-filled partials [like this](https://github.com/sds/haml-lint/commit/6ae3#diff-d889596ef7ca04c70befd0d1c7d5923ea034225e4a21f31b9fb41b649a1e6f0eR100-R106), so I've tweaked the spec to expect a different outcome on ruby 2.7.

sds:
Released in [0.37.0](https://github.com/sds/haml-lint/releases/tag/v0.37.0).

ghiculescu:
Thanks for adding this! We saw a similar 50% improvement in performance because of it. Even better when combined with https://github.com/sds/haml-lint/pull/322

PR #336, opened 2021-07-07
No linter include/exclude by default

### LINKED ISSUE #332: haml-lint ignores `.haml-lint_todo.yml`?
running gem 0.3.7

generate the todo with `haml-lint --auto-gen-config`

Added the new file to `.haml-lint.yml`

The ignored errors are still there.

PR where it happens: https://github.com/openSUSE/open-build-service/pull/11227/commits/7687b4e2f05a9aa76d6005be3d55e92f225c38f3

dswij:
Closes #332.

Currently every linter is ensured to have include and exclude configuration by default. This does not play nice when merging two configurations, since it overwrites the old one.

Another idea is to change the merge behavior instead, such that it only overwrites if the new array is not empty. However this will hinder the possibility to overwrite old include/exclude config with empty array.

### DIFF
--- a/lib/haml_lint/configuration.rb
+++ b/lib/haml_lint/configuration.rb
@@ -97,7 +97,7 @@ module HamlLint
     def validate
       ensure_exclude_option_array_exists
       ensure_linter_section_exists
-      ensure_linter_include_exclude_arrays_exist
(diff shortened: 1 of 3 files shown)


sds:
Thank you!

PR #337, opened 2021-07-21
Adds OffensecountReporter and unit tests

ireneybean:
This adds an offense count reporter in the style of rubocop's offense count formatter. It simply prints out the number of lints for each linter in descending order. For RuboCop lints, it will group by `"RuboCop: <offense name>"` which it extracts from the message. Thank you for your consideration :-)

### DIFF
--- a/lib/haml_lint/reporter/offensecount_reporter.rb
+++ b/lib/haml_lint/reporter/offensecount_reporter.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: false
+
+module HamlLint
+  # Outputs the a list of lints with a count of how many of each were found.
+  # Ordered by descending count
+  class Reporter::OffensecountReporter < Reporter
+    def display_report(report)
+      lints = report.lints
+      total_count = lints.count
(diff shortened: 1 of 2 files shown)


sds:
Apologies for the delay in response, @ireneybean.

Any reason we capitalized as `OffensecountReporter` and not `OffenseCountReporter`? Otherwise, this looks good to me.

ireneybean:
I started out with it as `OffenseCountReporter` but couldn't figure out how to get that to work from the command line because of line 65 in `options.rb`, which is where the reporter class is loaded... 

```
@options[:reporter] = load_reporter_class(reporter.capitalize)
```

I also tried to think of a nice easy one-word name but nothing was occurring to me.

If you have a suggestion or I'm missing something obvious, I'm more than willing to change it as it bothers me as well ;-)

sds:
Apologies for how long this took to merge. You can invoke this reporter with `--reporter offense-count`.

ireneybean:
Oh, yay! Thanks for addressing the multiword situation too!

PR #338, opened 2021-07-24
Fix inherits_from global configuration option to not be ignored

### LINKED ISSUE #311: LineLenght in todo are ignored with custom lenght in main
Hello

First, thank you for this gem, it helps us to keep our code clean.

I have an issue with LineLenght defined as follow:

```yaml
# .haml-lint.yml
---
inherits_from: .haml-lint_todo.yml

linters:
  LineLength:
    max: 120

# .haml-lint_todo.yml
linters:

  # Offense count: 3
  LineLength:
    exclude:
      - "app/views/devise/confirmations/new.html.haml"
      - "app/views/devise/registrations/edit.html.haml"
      - "app/views/devise/shared/_links.html.haml"
```

Given the configuration, we should expect these three files to be ignored by Haml-lint, but they are not:

```text
app/views/devise/confirmations/new.html.haml:7 [W] LineLength: Line is too long. [155/120]
app/views/devise/registrations/edit.html.haml:34 [W] LineLength: Line is too long. [127/120]
app/views/devise/shared/_links.html.haml:18 [W] LineLength: Line is too long. [137/120]
```

The issue occurs in Haml-lint `0.28` and `0.32`. I have not tested other versions.

Does that means the inherited configuration is overridden by the main configuration instead of being merged ?

yamotech:
The `smart_merge` method did not reflect the `child` configuration when merging keys with array values, so it has been changed to reflect both the `parent` and `child` configurations.

Fixes #311

### DIFF
--- a/lib/haml_lint/configuration.rb
+++ b/lib/haml_lint/configuration.rb
@@ -73,6 +73,8 @@ module HamlLint
         case old
         when Hash
           smart_merge(old, new)
+        when Array
+          old | new
         else
           new
         end
--- a/spec/haml_lint/configuration_spec.rb
+++ b/spec/haml_lint/configuration_spec.rb
@@ -81,7 +81,7 @@ describe HamlLint::Configuration do
     let(:config) { described_class.new(old_hash) }
     subject { config.merge(described_class.new(new_hash)) }
 
-    context 'when exclude is not explicitly declared on child configuration' do
+    context 'when `exclude` is not explicitly declared on child configuration' do
       let(:old_hash) do
         {
           'linters' => {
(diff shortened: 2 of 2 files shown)


sds:
Thanks for the pull request, @yamotech.

I understand this fixes a particular issue, but it's a fundamental change in the merge behavior that I'm not sure makes sense in all contexts.

Some custom hooks may have ordered lists for which a set union of the two merged lists may not make sense. If we can perhaps restrict this behavior in some way so that it is more predictable, that would be ideal.

sds:
Going to close due to age. Happy to discuss options if we think this is still worth pursuing. Thanks!

PR #340, opened 2021-09-14
Allow for hash attribute to be on separate lines

### LINKED ISSUE #339: Multiline hash attributes shouldn't get flagged for spacing issue
Haml Lint version: 0.37.1
Using the default value for `SpaceInsideHashAttributes`

Running on a file that looks like this:

```haml
%div
  %p{
    "data-stuff": "thing",
    "data-more": "double",
    "another": "fun"
  }
```

I would expect no errors, but I get the following:

```text
test.html.haml:2 [W] SpaceInsideHashAttributes: Hash attribute should start with one space after the opening brace
test.html.haml:2 [W] SpaceInsideHashAttributes: Hash attribute should end with one space before the closing brace
```

bfad:
When using hash attributes, the `SpaceInsideHashAttributes` should allow
the following to pass linting whether its value is set to 'spaces' or
'no_spaces':

    .container
      %my_tag{
        attr: 'one',
        next: 'attribtue',
        last: 'fun'
      }

This commit allows for that.

Fixes #339

### DIFF
--- a/lib/haml_lint/linter/space_inside_hash_attributes.rb
+++ b/lib/haml_lint/linter/space_inside_hash_attributes.rb
@@ -9,15 +9,15 @@ module HamlLint
     STYLE = {
       'no_space' => {
         start_regex: /\A\{[^ ]/,
-        end_regex: /[^ ]\}\z/,
+        end_regex: /(?:^\s*\}|[^ ]\})\z/,
         start_message: 'Hash attribute should start with no space after the opening brace',
-        end_message: 'Hash attribute should end with no space before the closing brace'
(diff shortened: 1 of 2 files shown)


sds:
Thank you!

sds:
Unfortunately, we had to revert this in #354. 

It turns out our CI was broken and not properly testing all versions. If you look into why this PR fails on some of the versions and fix it, happy to merge!

bfad:
@sds It looks like only older versions of HAML were failing, which makes sense since multi-line attributes were [added in 5.2](https://github.com/haml/haml/blob/main/CHANGELOG.md#521).

PR #343, opened 2021-11-27
Fix typo in linter doc for RubyComments

MaxLap:
"Space after `#` means comment is actually treated as Ruby code" is wrong.

2 options, I went with the first:

* Space before `#` ...
* Space after `-` ...


sds:
Thank you!

Issue #217, opened 2017-04-06
Autocorrect feature


Ana06:
Supporting autocorrect as Rubocop does would be great. :slightly_smiling_face:

michaelherold:
You're right, this would be a great feature. I consciously steered around that problem and haven't looked into it yet because, frankly, it scares me. 😅

stephen-puiszis:
@michaelherold what are the scary issues and how do we start addressing them?

sds:
RuboCop benefits from the hard work of the `parser` gem, which allows you to modify the syntax tree and re-serialize back to code without much effort.

There is no such parser for HAML. Thus HAML-Lint would effectively need to duplicate the HAML project to support its own parser that allows for mutating the syntax tree.

That is a huge amount of work. I would love to see someone step up to tackle it, but I'm skeptical the benefit is worth the development effort.

stephen-puiszis:
@sds thanks for the info, I assumed it would take quite a bit of work. I did some googling and found the [haml_parser gem](https://github.com/eagletmt/haml_parser). My intuition is that it needs a lot of work, but might provide a starting point. 

As someone working on a rescue project at the moment, I would love an auto-correct tool to help automate some of the cleanup process as the code is quite awful. I'm sure others would benefit from this as well... but you're right, it all eventually comes down to time and effort

Ana06:
@stephen-puiszis is the haml_parser gem really related. In the readme it says:

> Welcome to your new gem! In this directory, you'll find the files you need to be able to package up your Ruby library into a gem

Ana06:
@sds 

> Thus HAML-Lint would effectively need to duplicate the HAML project to support its own parser that allows for mutating the syntax tree.


Could the parser be useful for something else apart from the autocorrect option? Because then this would have more sense.

and anyway this parser should be an independent gem, not inside haml.lint, right?

sds:
Ideally this would be a contribution to the `haml` gem itself so that it benefits the most people. However, it's likely that there will be performance tradeoffs in having a parser that collects enough information to support syntax tree editing and code generation, so it would likely be a separate gem.

In either scenario: yes, this should not solely exist in the `haml_lint` gem.

sam0x17:
@michaelherold maybe a good initial step would be to add a --fix / --auto-correct option that only fixes rubocop portions of each haml file e.g. the ruby parts. I imagine that would be a lot easier?

Intrepidd:
[...]

PR #346, opened 2022-02-07
Enable RuboCop auto-correction

MaxLap:
So... I spent way more time than I wanted to on this. Folks from https://github.com/sds/haml-lint/issues/217 may be interested.

This draft PR adds auto-correction via RuboCop to Haml. It's not ready to be merged, but it's usable.

* Feature wise, it's pretty much finished.
* Test wise, there isn't much to add that I can think of.
* Code wise, there are some improvements that could be made... There always is.
* Documentation/comment wise, that was not my focus, so most of it is missing.

The system doesn't make a new parser for Haml. Here is a basic overview:
* Uses Haml's parser, but works around quirks by using the raw lines of interest from the haml file.
* First generates an array of "chunks" of ruby code that represent the Haml file.
* Not every chunks can be autocorrected, some are just placeholder for things like a "plain" to be able to correct the indentation and to avoid RuboCop thinking that an `if` is empty.
* The chunks are all written to a single file. Chunks that can be auto-corrected are wrapped with markers unique to the line in the file (ex: `haml_lint_marker_62`)
* Rubocop is then run onto that generated file, with a system to get the correct config.
* We then iterate over the chunks, find the corresponding new code using the chunks, and apply corrections back in the haml.

Can be used with:
* `-a` for safe auto-correct (which must be supported by your rubocop)
* `-A` for full auto-correct
* `--auto-correct-only` to ignore linters which don't support auto-correct (right now, only RuboCop does), and to ignore lint messages which were not auto-corrected (other than errors).
* There may be annoyances related to the Include/Exclude of your .rubocop.yml file. The linter writes a `Tempfile` with the `.rb` extension appended to your filename, so something like `....new.html.haml.rb`.

Note: Be careful just running this on your views and pushing this. You better test, because Rubocop has bugs, and I probably didn't catch every one in this feature.

I've tried to split some misc stuff into the early commits. And the last commit is all of the tests. (Something like 4k lines)

Context: I felt I was in a very good position to write this. I had access to a codebase with with 80000 lines of haml, written mostly by juniors with very little oversight.. it includes every code pattern you can think of. I could quickly hook the test system to store the generated html files. So I could compare the before/after to make sure the results were identical (except for the empty lines).

I've already spent more than 2 months working on this in my free time. Much more than I hoped. At this point, it works. I ran this on the whole codebase. A few files have patterns that are not supported (which are just skipped). This auto-corrects 25k lints (mostly about quotes and hash-rockets). The (incomplete but close to 1000) tests all
generated the exact same html (except for blank lines, which is expected).

My questions are:
* Is this feature likely to be accepted after it's been cleaned up and comments added?
* Is anyone willing to give a hand with finishing this?


config/default.yml
```diff
@@ -88,30 +88,94 @@ linters:
...
-      - Metrics/BlockNesting
-      - Metrics/LineLength
-      - Naming/FileName
-      - Style/FrozenStringLiteralComment
-      - Style/IfUnlessModifier
-      - Style/Next
-      - Style/WhileUntilModifier
+      # We allow the LineLength cop (by not having a Enabled: false for it),
+      # This way, other cops which rely on it can do so.
+      - <%= rubocop_version < '0.79' ? 'Metrics/LineLength' : 'Layout/LineLength' %>
```
sds, line 93: I would be in favor of requiring RuboCop >= 1.0 so we don't need to include workarounds like these.

config/default.yml
```diff
@@ -88,30 +88,94 @@ linters:
...
+
+    sent_to_rubocop:
```
sds, line 95: Can we add a comment explaining why this is needed versus a separate `.rubocop.haml-lint.yml` file?

Might make more sense here, but `sent_to_rubocop` is an odd setting name.

Gemfile
```diff
@@ -18,3 +18,5 @@ gem 'overcommit', '0.58.0'
 gem 'rubocop', '0.79.0'
 
 gem 'coveralls', require: false
+
+gem 'pry'
```
sds, line 22: Any hard reason we need this? `binding.irb` gets you most of the functionality without an additional dev dependency.

lib/haml_lint/linter.rb
```diff
@@ -7,6 +7,9 @@ module HamlLint
   class Linter
     include HamlVisitor
 
+    # Linters can override this to indicate that they support the autocorrect system
+    SUPPORTS_AUTOCORRECT = false
```
sds, line 11: Might make sense to make this work more like `attr_accessor` so that rather than overriding a constant we can simply add to a linter class implementation:

```ruby
supports_autocorrect true
```

...or similar.

sds (commented):
Thanks for the PR, @MaxLap.

There's a lot here to enable this functionality, and it's clear you've put a lot of effort into it. It's great that the autocorrect functionality is opt-in, so hopefully this doesn't affect anyone's workflow who has no interest in the implementation.

I would like to make clear that I personally don't have the bandwidth to support these changes, but I am happy to merge if you are open to addressing bugs as they arise. We can also clearly state in the README that the functionality is experimental for the time being.

In short, happy to merge if you are willing to take it the rest of the way. Thanks for showing off a proposal with working code.

Gemfile
MaxLap: Honestly, I've always used pry so I didn't know about `binding.irb`. I wasn't planning on leaving it there in the final commit anyways, but I'll take a look at `binding.irb`

config/default.yml
MaxLap: I have no problem extracting it to a file. Not being my gem, I tried not to change everything, and "config" for rubocop was already in that file (ignored lints), but I needed more granularity. If you prefer a dedicated file, it can be done.

config/default.yml
MaxLap: Alright, I'll do the switch then.

MaxLap:
Great, thanks for the positive answer and the code reviews. I have no problem supporting those changes.

I kind of burned myself out spending months on this already, I'll work on it when the spark come back a little.

Upgrading RuboCop is nice and I'm interested, but that also means new offenses in existing code. The most painful part of Rubocop is dealing with the `Metrics` cops. AbcSize, MethodLength, CyclomaticComplexity, PerceivedComplexity, ClassLength, etc. all feel painful, and I'm always wondering if it will be ok if I disable it for method X, Y, Z. I don't feel those cops add much but development overhead. Would you be open to me disabling them all entirely?

One thing I forgot to mention in the initial description is that if a file has lots of auto-corrections, the lints end up being in kind of meaningless locations, because as the Ruby file gets changed, RuboCop reports line numbers according to "after" the previous changes. So the basic mapping we use kind of gets broken. I have no idea if there is a solution to that and, to me, improving that is very low priority.

arrowcircle:
Thanks for your work! Hope this significant thing will be merged soon

MaxLap:
Geez, it's already been 4 months. I do want to get back to this, but work, vacation and other projects have been leaving me with little time. Just wanted to show a sign of life on my side :)

bloycey:
Unfortunately I don't have the skills to personally contribute to this PR but just wanted to say thank you for all your work on this so far. Really hoping it gets merged as it would be a huge time saver!

MaxLap:
So, finally gotten around to finishing this. Lots of cleanups and improvements to overall usability of the feature. At this point, I believe the feature is ready for a merge or a more thorough review.

Again, I split the commits so that they are self contained. Most of the first ones are just small improvements to Haml-Lint that felt needed or were found while developing the feature. The feature itself is in the 1 of the last 3 commits, and the tests are in another one of the last 3 commits.. Those are big.

Other than reviewing, here are some questions I had:
* Which appraisals should we use? Right now, I added one for Rubocop 1.0, 1.25 and 1.50. Those all test with the same HAML version. It can be changed pretty easily.
* RuboCop supports the terms `auto-correct` and `autocorrect` in its options, but encourages `autocorrect`. When I made the feature, I don't think that was the case. Either way, should we support both or only one? Thoughts? Right now, the PR supports `auto-correct` only because I wanted to wait for your input.

Finally, I did not resolve every Rubocop offenses because:
* Not sure which version you want (I recommend going for the latest)
* Some changes may be things you would prefer to disable
* There are lots of new cops to choose to enable/disable
* The changes affect basically every file in the project and this PR feels big enough
* I have a hard time caring about all the Metrics cops

I did upgrade RuboCop to 1.0 since that's the minimum, but only applied most fixes to the code I'm adding without editing the rest of the project.

MaxLap:
Force pushed a fix for the Rubocop config_file option, which failed when given a relative path. (All tests used absolute path, ops!) 
Thanks to @thomasbrus for finding, investigating and reporting the bug.

If anyone want to try this feature and see how it goes, it would be very helpful to get some feedback on this.

To install this the gem with the feature, add this to your Gemfile:
`gem 'haml_lint', github: "maxlap/haml-lint", branch: 'autocorrect_rubocopv2', require: false`

Then, you can run with autocorrection using either `haml-lint -a --auto-correct-only` or `haml-lint -A --auto-correct-only` depending on your preference based on RuboCop's safe vs unsafe autocorrection.

The `--auto-correct-only` makes it faster for the first use by skipping linters which cannot do autocorrection.

Make sure to run your tests afterward! This can be a lot of changes depending on the code-base and your rubocop configuration.

Any feedback is welcome, either here or to the email in my profile.

baelter:
Works like a charm!

You should make up your mid about quotes and autocorrect that
> gem 'haml_lint', github: "maxlap/haml-lint", branch: 'autocorrect_rubocopv2', require: false
:)

MaxLap:
Haha @baelter, good catch 😄 I fixed it.

And thanks for the feedback that you had no issues!

sds:
Sorry for the delay in response, @MaxLap. Appreciate your effort on this. Responses inline.

> * Which appraisals should we use? Right now, I added one for Rubocop 1.0, 1.25 and 1.50. Those all test with the same HAML version. It can be changed pretty easily.

If there is a significant difference in the AST exposed by RuboCop such that an appraisal version for it is necessary, we can add one for each substantial change. If not, I would suggest not adding appraisals for RuboCop, and simply require RuboCop 1.x as a minimum requirement.

> * RuboCop supports the terms `auto-correct` and `autocorrect` in its options, but encourages `autocorrect`. When I made the feature, I don't think that was the case. Either way, should we support both or only one? Thoughts? Right now, the PR supports `auto-correct` only because I wanted to wait for your input.

Comfortable with `auto-correct` only for now.

> Finally, I did not resolve every Rubocop offenses because:
> 
> * Not sure which version you want (I recommend going for the latest)

I'm fine with that as well.

> * Some changes may be things you would prefer to disable
> * There are lots of new cops to choose to enable/disable

I'm supportive of disabling by default or accepting whatever autocorrections RuboCop implements—whichever is easier.

> * The changes affect basically every file in the project and this PR feels big enough
> * I have a hard time caring about all the Metrics cops

Fine with disabling the Metrics-related cops for the files you've added or methods you've modified.

> I did upgrade RuboCop to 1.0 since that's the minimum, but only applied most fixes to the code I'm adding without editing the rest of the project.

Sounds great, thank you.

---

Overall I'm supportive of this change, just need the build to be green. I know this has been sitting around for a while (apologies again) but happy to merge and cut a release once the build is green. Thank you!

README.md
```diff
@@ -401,6 +416,12 @@ appraisal bundle install
 appraisal bundle exec rspec
 ```
 
+To run the test suite faster (about 4x), you can skip the RuboCop invocations for testing
```
ghiculescu, line 419: I'm not a haml-lint maintainer, I'm just some guy, so take this with a grain of salt.

Given that [tests take 9s to run](https://github.com/sds/haml-lint/actions/runs/5248634051/jobs/9480384913?pr=346), I am not sure it's worth the extra complexity of this feature.

README.md
```diff
@@ -91,6 +98,14 @@ Command Line Flag          | Description
 `-v`/`--version`           | Show `haml-lint` version
 `-V`/`--verbose-version`   | Show `haml-lint`, `haml`, and `ruby` version information
 
+## Autocorrection
+
+`haml-lint` support automatically correcting some of the errors that it detects, this includes
```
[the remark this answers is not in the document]
MaxLap: Thnx!

README.md
MaxLap: It's a simple feature that has no code in the lib (It's basically a single stub in the tests). For me, it's the difference 17s and 5s. It can matter when working locally.

So for now I'll leave that in, but thanks for your input!

MaxLap:
@sds Alright, everything is green.

I switched to Rubocop 1.50.2 since that's the last one supporting Ruby 2.6.

I left a single Appraisal for Rubocop 1.0. as a just in case.

I applied all of the auto-correction of Rubocop, disable a few that I felt didn't add anything (ex: missing call to `super`). And added some rubocop:disable for metrics related stuff.

sds:
Thanks!

PR #348, opened 2022-02-24
Migrate from Travis to GitHub Actions

sds:
It's time.


lib/haml_lint/linter/rubocop.rb
```diff
@@ -27,8 +27,6 @@ def visit_root(_node)
       find_lints(extracted_source.source, extracted_source.source_map)
     end
 
-    private
```
sds, line 30: There was a RuboCop warning about this `private` not applying to the `self.rubocop_cli` class method below. We can't actually make this method private due to the number of times we reference it in tests, so just move this down to address the RuboCop warning.

PR #358, opened 2022-02-26
Allow for hash attribute to be on separate lines

### LINKED ISSUE #357: Multiline hash attributes shouldn't get flagged for spacing issue
Re-opening #339 because the fix in #340 was reverted.

Haml Lint version: 0.37.1
Using the default value for `SpaceInsideHashAttributes`

Running on a file that looks like this:

```haml
%div
  %p{
    "data-stuff": "thing",
    "data-more": "double",
    "another": "fun"
  }
```

I would expect no errors, but I get the following:

```text
test.html.haml:2 [W] SpaceInsideHashAttributes: Hash attribute should start with one space after the opening brace
test.html.haml:2 [W] SpaceInsideHashAttributes: Hash attribute should end with one space before the closing brace
```

bfad:
When using hash attributes, the `SpaceInsideHashAttributes` should allow
the following to pass linting whether its value is set to 'spaces' or
'no_spaces':

    .container
      %my_tag{
        attr: 'one',
        next: 'attribtue',
        last: 'fun'
      }

This commit allows for that.

Fixes #357

### DIFF
--- a/lib/haml_lint/linter/space_inside_hash_attributes.rb
+++ b/lib/haml_lint/linter/space_inside_hash_attributes.rb
@@ -9,15 +9,15 @@ module HamlLint
     STYLE = {
       'no_space' => {
         start_regex: /\A\{[^ ]/,
-        end_regex: /[^ ]\}\z/,
+        end_regex: /(?:^\s*\}|[^ ]\})\z/,
         start_message: 'Hash attribute should start with no space after the opening brace',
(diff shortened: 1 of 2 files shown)


bfad:
@sds I updated my tests to only run on versions that support multiline attributes.

sds:
Thanks for the fix! Released [0.40.0](https://github.com/sds/haml-lint/releases/tag/v0.40.0) to re-include the fix.

PR #362, opened 2022-07-28
fix: use unsafe_load for ruby 3.1

Keulsss:


SeanLF:
This fixes #363

sds:
Considering this was opened a while ago and the failing tests haven't been addressed, I'm going to close this PR as abandoned. 

Would welcome a change that solves the issue if there is appetite for it.

Issue #341, opened 2021-11-08
`InstanceVariables` doesn't work with if statements


ghiculescu:
This test, if added [here](https://github.com/sds/haml-lint/blob/master/spec/haml_lint/linter/instance_variables_spec.rb), fails

```ruby
      context 'in an if statement' do
        let(:haml) { <<-HAML }
          - if @greeting
            Hello
        HAML

        it { should report_lint line: 1 }
      end
```

With this output:

```
  1) HamlLint::Linter::InstanceVariables when the file name matches the matcher and there is an instance variable in an if statement is expected to report a lint on line 1
     Failure/Error: it { should report_lint line: 1 }
       expected that a lint would be reported on line 1
     # ./spec/haml_lint/linter/instance_variables_spec.rb:59:in `block (5 levels) in <top (required)>'
```

ghiculescu:
This is because `HamlLint::Tree::SilentScriptNode#parsed_script` is returning `nil`, I guess because `"if @greeting"` is an incomplete Ruby statement 🤔

sds:
Fixed in #364.

PR #364, opened 2022-08-11
Fix InstanceVariables check for multiline blocks

### LINKED ISSUE #245: InstanceVariables check not working for multi-line blocks
The following generates an `InstanceVariables: Avoid using instance variables in partials views` warning, as expected:
```haml
= simple_form_for(@widget)
```
as does
```
= simple_form_for(@widget) { |f| f.input :name }
```
But this doesn't:
```haml
= simple_form_for(@widget) do |f|
  = f.input :name
```

Using haml-lint 0.24.0

I haven't yet tried on any other versions.

glennakamura:
Add surrounding statements and keywords to ruby script fragments
so parsed_script can return a valid abstract syntax tree.

Since parsed_script is only used by the InstanceVariable check it should
be sufficient to augment the script fragement with minimal dummy statements
to generate valid ruby syntax for the parser.  The returned AST may not
represent the actual document structure but it's suitable for detecting
usage of instance variables within the script fragment.

Append ';end' to script nodes with children to close their blocks.
As a special case we append ';when 0;end' to close case statements.
Prepend appropriate statements to script nodes with midblock keywords:
else, elsif, when, rescue, ensure

Fixes #245 #341

### DIFF
--- a/lib/haml_lint/tree/script_node.rb
+++ b/lib/haml_lint/tree/script_node.rb
@@ -9,7 +9,8 @@ module HamlLint::Tree
     #
     # @return [ParsedRuby] syntax tree in the form returned by Parser gem
     def parsed_script
-      HamlLint::ParsedRuby.new(HamlLint::RubyParser.new.parse(script))
+      statement = children.empty? ? script : script + (@value[:keyword] == 'case' ? ';when 0;end' : ';end')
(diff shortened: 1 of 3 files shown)


glennakamura:
Fixes #245 #341

sds:
Thanks for the PR and for the thorough tests! Fixed the lints in #365.

PR #367, opened 2022-08-21
Compatibility with Haml 6

navidemad:
The API for handling error changed and added Haml::HamlError (https://github.com/haml/haml/search?q=unify)
The parser of the new haml version seems to return a class Error instead of throwing, so I patched it with a monkey patch while i'm waiting for news in the following issue : https://github.com/haml/haml/issues/1082

The issue got resolved, and now it's throwing exception correctly.


sds:
Thank you!

Issue #361, opened 2022-07-26
Avoid joining multiline Ruby to not anger RuboCop


danieldiekmeier:
Thanks for building HAML-Lint, it really helped me clean up the templates in our Rails app!

I have a file where I want to break inside a hash to make my file more readable. Simplified example:

```haml
.example{class: class_names({
  optional: maybe?
})}
```

This works and is valid Haml, but my RuboCop configuration is set to `Layout/SpaceInsideHashLiteralBraces: no_space`, which makes RuboCop angry because  HAML-Lint joins the lines of the hash but keeps spaces around the hash braces.

To make sure I understood this right, I added a test case to `spec/haml_lint/ruby_extractor_spec.rb`:

```ruby
context 'with a tag with hash attributes containing a hash' do
  let(:haml) { <<-HAML }
    %tag{class: class_names({
      optional: maybe?
    })}
  HAML

  its(:source) { should == normalize_indent(<<-RUBY).rstrip }
    {}.merge(class: css_names({
      optional: maybe?
    }))
    _haml_lint_puts_0 # tag
    _haml_lint_puts_1 # tag/
  RUBY
end
```

And indeed, the diff is:

```diff
-{}.merge(class: css_names({
-  optional: maybe?
-}))
+{}.merge(class: class_names({ optional: maybe? }))
 _haml_lint_puts_0 # tag
 _haml_lint_puts_1 # tag/
```

Of course, if we were to simply remove the spaces, _my_ RuboCop would be happier, but if RuboCop was configured the other way (`space`), it would be angry as well.

My suggestion would be to _not_ join the lines of a multiline hash. Indentation [already does not matter ](https://github.com/sds/haml-lint/blob/1231b394bcc2b6c9c47a7a18074b43b258a0872e/config/default.yml#L90), so I think this could be an option?

The same change would probably need to be made for **parentheses**, because RuboCop dislikes this HAML for the same reason:

```haml
.example{class: class_names(
  {optional: maybe?}
)}
```

I could give this a go if you are open to the change. Thanks for your consideration!

sds:
Very open to a contribution that addresses! Thanks for the detailed explanation.

danieldiekmeier:
Closing this because I created a PR which was merged! Thanks again!

PR #369, opened 2022-08-25
Keep newlines in extracted Ruby

danieldiekmeier:
PR based on my suggestion in #361.

This basically stops HamlLint from removing the newlines from extracted Ruby code.

Before, 

```haml
%tag{class: some_method({
  one: 1,
  two: 2
})}
```

got extracted as:

```ruby
{}.merge(class: some_method({ one: 1, two: 2 })}
```

but this made RuboCop angry if you configured it to disallow spaces around the hash braces.

Now, it will be extracted as:

```ruby
{}.merge(class: some_method({
one: 1,
two: 2
}))
```

I also had to add `Layout/FirstHashElementIndentation` as a default ignored RuboCop Cop because  the indentation of the hash elements gets lost in the extraction. I think this is fine, because `Layout/HashAlignment` is already disabled.

### DIFF
--- a/spec/haml_lint/ruby_extractor_spec.rb
+++ b/spec/haml_lint/ruby_extractor_spec.rb
@@ -234,38 +234,41 @@ describe HamlLint::RubyExtractor do
       its(:source_map) { should == { 1 => 1, 2 => 1, 3 => 1, 4 => 1, 5 => 1 } }
     end
 
-    context 'with a tag with hash attributes containing a hash with newlines' do
-      let(:haml) { <<-HAML }
-        %tag{class: some_method({
+    # Multiline attributes were introduced in 5.2.1
+    if Haml::VERSION >= '5.2.1'
+      context 'with a tag with hash attributes containing a hash with newlines' do
+        let(:haml) { <<-HAML }
+          %tag{class: some_method({
+            one: 1,
+            two: 2
+          })}
+        HAML
+
+        its(:source) { should == normalize_indent(<<-RUBY).rstrip }
+          {}.merge(class: some_method({
           one: 1,
           two: 2
-        })}
-      HAML
+          }))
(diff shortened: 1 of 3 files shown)


sds:
Not sure if you received a notification about test failures, @danieldiekmeier. Happy to merge this functionality if we get a passing test suite. Thank you.

danieldiekmeier:
Thanks for the reminder. I didn't realize the error was on my end – it had to do with multiline attributes in Haml < 5.2.1. I adopted the conditional testing from `space_inside_hash_attributes_spec.rb` and now it works!

spec/haml_lint/ruby_extractor_spec.rb
```diff
@@ -224,12 +224,51 @@
...
         _haml_lint_puts_0 # tag
         _haml_lint_puts_1 # tag/
       RUBY
 
-      its(:source_map) { should == { 1 => 1, 2 => 1, 3 => 1 } }
+      its(:source_map) { should == { 1 => 1, 2 => 1, 3 => 1, 4 => 1, 5 => 1 } }
+    end
+
+    # Multiline attributes were introduced in 5.2.1
+    if Haml::VERSION >= '5.2.1'
```
sds, line 238: Just to confirm: should this conditional be in the linter itself, not in the tests? Otherwise anyone running the linter with a version of HAML would see the test failures this is attempting to correct, or am I misunderstanding the conditions where this occurs?

danieldiekmeier: I think this is the same case like in `space_inside_hash_attributes_spec.rb` – Any case where this linter could activate would already be invalid in old versions of HAML. But if you‘d rather have it moved, I can do that!

danieldiekmeier:
Thank you!

PR #370, opened 2022-08-25
Add TruffleRuby to CI

gogainda:


sds:
Can we provide an explanation why it is worthwhile to be testing _all_ PRs against a potentially unstable (a.k.a. "head") version of a third-party Ruby implementation?

Given this gem is written in vanilla Ruby, it's not clear what the value is in testing against additional Ruby implementations. If HAML-Lint does not work on other Ruby implementations, that is the responsibility of those implementations, not HAML-Lint.

If I'm missing some context that explains the need for this, appreciate you including it in the PR description. Thank you.

PR #383, opened 2022-12-25
Allow inheriting from relative paths to the home directory

gonubana:
by always expanding paths before use.


sds:
Going to close since the tests are failing. Feel free to reopen with a working PR. Thanks!

PR #389, opened 2023-03-15
ci: Update actions to latest with Dependabot

nschonni:
Notice a few were old, so manually bumped them and added Dependabot to keep dependencies up-to-date


sds:
Thanks!

PR #397, opened 2023-03-16
chore: Resync gemfiles with Appraisal

nschonni:
Maybe Dependabot should also be added to point at these Gemfiles, or some of these deps removed from the Gemfile to the gemspec to reduce the noise


sds:
Thanks!

PR #403, opened 2023-05-20
Add NoPlaceholders linter

### LINKED ISSUE #371: Rule Idea: no placeholders
Using the `placeholder` attribute [is considered an anti-pattern](https://www.smashingmagazine.com/2018/06/placeholder-attribute/) as it has a lot of accessibility drawbacks. It would be great if we could enforce not using this in our haml files.

benmelz:
Implements a `NoPlaceholders` linter to ensure placeholder attributes are not used. It is supported by regex matching against the source hash/html attributes that searches for `placeholder` keys.

Closes #371

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -76,6 +76,9 @@ linters:
   MultilineScript:
     enabled: true
 
+  NoPlaceholders:
+    enabled: false
+
   ObjectReferenceAttributes:
     enabled: true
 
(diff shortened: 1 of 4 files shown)


config/default.yml
```diff
@@ -76,6 +76,9 @@ linters:
   MultilineScript:
     enabled: true
 
+  NoPlaceholders:
+    enabled: false
+
```
benmelz, line 81: I'm basing this decision mostly on the fact that `AltText` (another HTML accessibility rule) has `enabled: false`.

lib/haml_lint/linter/README.md
```diff
@@ -501,6 +502,24 @@ will compile and run:
...
+**Bad**
+```haml
+%input{ placeholder: 'Placeholders arent very accessible' }
+```
+
+**Good**
+```haml
+#my-details Placeholders arent very accessible
+%input{ 'aria-describedby': 'my-details' }
+```
```
benmelz, line 518: I wasn't really sure what to include here, so I defaulted to the most common alternative. Open to any other suggestions.

PR #418, opened 2023-07-07
Improve coverage related to new Rubocop system

MaxLap:
I noticed some files are very bad coverage. Turns out there was a big unused method, and the Runner code related for auto-correction wasn't tested at all.


sds:
Thanks!

PR #423, opened 2023-07-13
Suppress compliancy warning of ruby parser while requiring

ItsEcholot:
This fixes the following warning when running haml-lint:
```
warning: parser/current is loading parser/ruby31, which recognizes 3.1.3-compliant syntax, but you are running 3.1.2.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

The approach has been copied from the `rubocop` gem (https://github.com/marcandre/rubocop-ast/commit/7081f7cae0326150b648c54a99e124998ab2bffb).


sds:
Thanks!

Issue #446, opened 2023-10-01
Add support to Haml 6.2


tagliala:
It is pretty unexpected that a test/development linter prevents the update of the production library.

I would prefer the linter to break, but that should not prevent the upgrade of a production library

I'm going to submit a PR to allow any version of Haml, but we can easily add back a minor or major release 

If haml lint is not compatible with the current version of Haml, it will fail with `HamlLint::Exceptions::UnknownHamlVersion`, so a new issue can be opened on this repo to ask for support

Issue #445, opened 2023-10-01
Failing spec on main branch


tagliala:
macOS 14.0
Ruby 3.2 (and 3.1)
Commit: ccdaed5

```
Failures:

  1) HamlLint::Linter::RuboCop autocorrect edge cases correct multiline brace layout of new_line with right indentation in silent script is expected to eq "%tag\n  - a = render 'something',\n               locals: { foo: bar,\n                         spam: 'more',\n    }\n"
     Failure/Error:
       document.source.should(
         matcher,
         -> { "Final HAML is different from expected. #{matcher.failure_message}\n#{format_lints}" }
       )
     
       Final HAML is different from expected. 
       expected: "%tag\n  - a = render 'something',\n               locals: { foo: bar,\n                         spam: 'more',\n    }\n"
            got: "%tag\n  - render 'something',\n           locals: { foo: bar,\n                     spam: 'more',\n    }\n"
     
       (compared using ==)
     
       2:Lint/UselessAssignment: Useless assignment to variable - `a`.
       3:Layout/ArgumentAlignment: Align the arguments of a method call if they span more than one line.
       3:Layout/ArgumentAlignment: Align the arguments of a method call if they span more than one line.
       3:Layout/SpaceInsideHashLiteralBraces: Space inside { missing.
       4:Layout/HashAlignment: Align the keys of a hash literal if they span more than one line.
       4:Layout/SpaceAfterComma: Space missing after comma.
       4:Layout/MultilineHashBraceLayout: Closing hash brace must be on the line after the last hash element.
       4:Layout/SpaceInsideHashLiteralBraces: Space inside } missing.
       4:Layout/TrailingWhitespace: Trailing whitespace detected.
       Diff:
       @@ -1,6 +1,6 @@
        %tag
       -  - a = render 'something',
       -               locals: { foo: bar,
       -                         spam: 'more',
       +  - render 'something',
       +           locals: { foo: bar,
       +                     spam: 'more',
            }
       
     # ./spec/haml_lint/linter/rubocop_autocorrect_edge_cases_spec.rb:218:in `block (4 levels) in <top (required)>'

  2) HamlLint::Linter::RuboCop autocorrect (script_examples.txt:484) Handles a for loop is expected to eq "haml_lint_marker_1\nfor value in list # rubocop:disable Style/For\n  HL.out = foo(bar: 123)\nend\nhaml_lint_marker_5\n"
     Failure/Error:
       subject.last_new_ruby_source.should(
         matcher,
[...]

PR #447, opened 2023-10-01
Allow Haml > 6.1

tagliala:
[Add Haml 6.2 compatibility](https://github.com/sds/haml-lint/pull/447/commits/437857a4edc369ba617936ed066d30e33a6eb23d) 

Remove upper version constraint to allow any version of Haml.

Haml-lint will break if the loaded Haml version is not supported, but this
change will allow developers to update the production library and decide how
to take action on the linter later.

Also fixes 6.1 appraisal file.

Close https://github.com/sds/haml-lint/issues/446

---

Also bumps rubocop and removes an outdated comment


tagliala:
Failures do not depend by these changes, see #445

lib/haml_lint/adapter.rb
```diff
@@ -21,7 +21,7 @@ def self.detect_class
       case version
       when '~> 4.0' then HamlLint::Adapter::Haml4
       when '~> 5.0', '~> 5.1', '~> 5.2' then HamlLint::Adapter::Haml5
-      when '~> 6.0', '~> 6.0.a', '~> 6.1' then HamlLint::Adapter::Haml6
+      when '~> 6.0', '~> 6.0.a', '~> 6.1', '~> 6.2' then HamlLint::Adapter::Haml6
```
tagliala, line 24: should we change `haml_version` to `haml_major_version` and go for `segments.first.to_i`?

```
      when 4 then HamlLint::Adapter::Haml4
      when 5 then HamlLint::Adapter::Haml5
      when 6 then HamlLint::Adapter::Haml6
```

PR #453, opened 2023-10-24
Fix markdown highlighting in linter readme

mjankowski:
Trailing asterisk was missing, leading to incorrect bolding of the snake_case rule description in the IdNames section.


sds:
Thanks!

Issue #452, opened 2023-10-24
Add `-s`/`--stdin` option support


benmelz:
To preface, I'm prepared to submit a PR for this myself, but I just wanted to get a feel for if it would be accepted/get an opinion on implementation before I put the work in.

<hr />

I would like to propose adding an option similar to rubocop's `-s` flag:

```
rubocop --help

...
-s, --stdin FILE                 Pipe source from STDIN, using FILE in offense
...
```

i.e. allow piping of source code directly into the command (along with a "fake" file name for reporters to use).

It would be very helpful for editor integrations, such as [the plugin for Jetbrains IDEs that I maintain](https://github.com/benmelz/haml-lint-intellij-plugin). For example a user reported issues with docker that would be [fixable by adding this flag](https://github.com/benmelz/haml-lint-intellij-plugin/issues/50#issuecomment-1777282129).

<hr />

As for an implementation, from what I've read it's tricky because the `HamlLint::Runner` seems pretty centered around running against files. I was thinking that it might be easy to add a condition such that when the flag is given, the method below reads from stdin and writes/returns a tempfile to target.

https://github.com/sds/haml-lint/blob/0d3151ead87be43abcda10ad2d75b1feb81efa66/lib/haml_lint/runner.rb#L134-L140

sds:
Open to a PR! [This feature was implemented in another project](https://github.com/sds/slim-lint/pull/125) if you wanted to crib from that, though it's possible these two projects are different enough that it's not 1:1.

MaxLap:
By the way, since we also have auto-correction now, you might end up needing a --stderr too like RuboCop has.
https://docs.rubocop.org/rubocop/1.57/usage/basic_usage.html

PR #454, opened 2023-10-28
Add `--stderr` option

benmelz:
As suggested by @MaxLap on #452 

This option can stand on its own and is pretty straightforward (just need to point the logger to `$stderr` I believe), so I figured I would implement it before I work out `--stdin`.

### DIFF
--- a/spec/haml_lint/cli_spec.rb
+++ b/spec/haml_lint/cli_spec.rb
@@ -38,6 +38,27 @@ describe HamlLint::CLI do
       end
     end
 
+    context 'when a logger is not passed' do
+      let(:logger) { nil }
+
+      it 'sets a default logger that writes to stdout' do
+        $stdout = StringIO.new
+        subject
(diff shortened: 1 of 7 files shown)


spec/haml_lint/cli_spec.rb
```diff
@@ -38,6 +38,27 @@
...
+    context 'when passed the --stderr flag' do
+      let(:logger) { nil }
+      let(:args) { %w[--stderr] }
+
+      it 'sets a logger that writes to stderr' do
+        $stderr = StringIO.new
+        subject
+        expect(cli.send(:log).instance_variable_get(:'@out')).to eq($stderr)
+      end
+    end
```
benmelz, line 60: These are preeeetty gross, but they're as direct as I could make them without exposing parts of `CLI` and `Logger` that don't need to be exposed.

The `$stdout/err = StringIO.new` keeps these test runs from cluttering up the rspec output.

sds:
Thanks!

PR #459, opened 2023-12-11
Allow Haml 6.3

### LINKED ISSUE #458: Allow Haml 6.3

tagliala:
Close #458


lib/haml_lint/adapter.rb
```diff
@@ -21,7 +21,7 @@ def self.detect_class
       case version
       when '~> 4.0' then HamlLint::Adapter::Haml4
       when '~> 5.0', '~> 5.1', '~> 5.2' then HamlLint::Adapter::Haml5
-      when '~> 6.0', '~> 6.0.a', '~> 6.1', '~> 6.2' then HamlLint::Adapter::Haml6
+      when '~> 6.0', '~> 6.0.a', '~> 6.1', '~> 6.2', '~> 6.3' then HamlLint::Adapter::Haml6
```
tagliala, line 24: should we change `haml_version` to `haml_major_version`, return `segments.first.to_i`, and go for:

```rb
      when 4 then HamlLint::Adapter::Haml4
      when 5 then HamlLint::Adapter::Haml5
      when 6 then HamlLint::Adapter::Haml6
```

or dynamic adapter resolution?

sds: I'm supportive of dropping older versions of HAML, especially v4. I don't see a legitimate reason to continue supporting such an old version in light of how long v6 has been available.

PR #469, opened 2024-01-06
Support for conditional comments in RuboCop linter

### LINKED ISSUE #468: Issue with Rubocop `Layout/IndentationConsistency` interaction with haml comment blocks
Given an html layout like:

```
!!!
%html
  %head
    /[if mso]
      <xml></xml>
  %body
```

Running haml-lint with rubocop enabled generates a "RuboCop: Layout/IndentationConsistency: Inconsistent indentation detected" failure.

The linter seems to want the contents of the html comment to be aligned with the start of the comment block -- I'm guessing that in the ruby<>haml translation something is lost here. Longer context/description: https://github.com/mastodon/mastodon/pull/28416#issuecomment-1879090547

Is there an obvious workaround or better haml approach here? Am I understanding what's happening? Would love guidance on best approach here (is this a haml lint bug? should we just disable that specific rubocop rule? etc)

MaxLap:
HAML is full of surprises features haha!

Fixes #468


mjankowski:
I don't fully understand the spec file format/approach here, but is it the case that with this fix it would solve the false positive cop in the original issue, while still enforcing valid format for the haml within the html comment?

MaxLap:
Yeah, the format of these tests is quite dense. 

Yes, the content nested under the conditional comment is still going to be formatted.

mjankowski:
> Yes, the content nested under the conditional comment is still going to be linted/formatted by RuboCop.

Excellent.

FYI - I updated the WIP branch where we saw this issue to point to this PR for haml-lint gem, and it does indeed resolve all the false positives we had around conditional comments.

PR #472, opened 2024-01-07
Test against Ruby 3.3

tagliala:


tagliala:
Ruby 3.3 with latest haml versions should also be "required"

sds:
Thank you.

PR #478, opened 2024-01-20
Add `-s`/`--stdin` option

### LINKED ISSUE #452: Add `-s`/`--stdin` option support

benmelz:
Closes #452

It's been a while, but finally had some bandwidth to write this out. As laid out in the issue, implements `-s`/`--stdin` to mirror the corresponding `rubocop` options.

The biggest obstacle was separating a source file's contents from its path in the context of `HamlLint::Runner`. I achieved that here by adding a lightweight wrapper class `HamlLint::Source` that accepts an IO object and path string separately, the former allowing for easy interchange between `StringIO` (from stdin) and `File` (from normal execution) objects.

I initially was passing around hashes (i.e. `{ io: ..., path: ... }`) but settled on a class as it's a bit more elegant and way nicer to document. I can also see the `Source` class taking on a little more repsonsibility with future refactoring as well (move `extract_applicable_files/sources` to a class method there, pass it directly to `Document`, etc).

Anyways, beyond that it was just little tweaks to get specs/autocorrect to play nice with the new abstraction/option respectively.

### DIFF
--- a/lib/haml_lint/options.rb
+++ b/lib/haml_lint/options.rb
@@ -70,7 +70,13 @@ module HamlLint
         @options[:autocorrect] ||= :safe
       end
 
-      parser.on('--stderr', 'Write all output to stderr') do
+      parser.on('-s', '--stdin FILE', 'Pipe source from STDIN, using FILE in ' \
+          'offense when combined with --auto-correct and --stdin.') do |file_path|
+        @options[:stdin] = file_path
+      end
+
+      parser.on('--stderr', 'Write all output to stderr except for the autocorrected source. ' \
+          'This is especially useful when combined with --auto-correct and --stdin') do
         @options[:stderr] = true
       end
     end
--- a/lib/haml_lint/options.rb
+++ b/lib/haml_lint/options.rb
@@ -32,7 +32,7 @@ module HamlLint
 
     private
 
-    def add_linter_options(parser) # rubocop:disable Metrics/MethodLength
+    def add_linter_options(parser) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       parser.on('--auto-gen-config', 'Generate a configuration file acting as a TODO list') do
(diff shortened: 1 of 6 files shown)


lib/haml_lint/options.rb
```diff
@@ -32,7 +32,7 @@ def parse(args)
 
     private
 
-    def add_linter_options(parser) # rubocop:disable Metrics/MethodLength
+    def add_linter_options(parser) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
benmelz, line 35: Can roll this back and instead try to break this method up, but I figured if we're already disabling `MethodLength`, this would be appropriate to disable as well.

sds (approved):
Thank you!

PR #487, opened 2024-02-02
Fix tag attribute spacing

### LINKED ISSUE #485: Layout/ArgumentAlignment - False positive?
I have cobbled together the following example:
```haml
.regular.html{
  'data-very-long-so-this-cant-be-on-the-line-above': 'something',
  'data-it-needs-at-least-two-attributes-to-fail': 'something'
}
```

According to the [Reference](https://haml.info/docs/yardoc/file.REFERENCE.html#attributes) this should be possible, but unfortunately running `haml-lint` I get the following errors:
```
test.haml:2 [W] RuboCop: Layout/ArgumentAlignment: Align the arguments of a method call if they span more than one line.
test.haml:2 [W] RuboCop: Layout/HashAlignment: Align the keys of a hash literal if they span more than one line.
```
Is this a bug? Is my example considered bad style? Is there a way around without putting the second line behind the first one?

With just one data attribute it works just fine:
```haml
.regular.html{
  'data-very-long-so-this-cant-be-on-the-line-above': 'something'
}
```
I do know that I could just do
```haml
.regular.html{ 'data-very-long-so-this-cant-be-on-the-line-above': 'something',
               'data-it-needs-at-least-two-attributes': 'something' }
```
but I want to enforce the LineLength cop, so this way isn't feasible in my case.

MaxLap:
Fixes #485

A side benefit of this change is that we generate less Ruby for RuboCop, so maybe a bit of performance benefit.


sds:
Thanks for the fix!

PR #492, opened 2024-02-12
Add GitHub Reporter

joergschiller:
We're using HAML-Lint together with RuboCop and Stylelint in a GitHub action to run linters as checks on pull requests. There already is [a RuboCop formatter](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/formatter/github_actions_formatter.rb) that emits GitHub workflow commands to annotate the source code with warnings and errors.

I've created a reporter based on that formatter. I tested it with a few warnings that occured within our codebase after your latest release.

image

Let me know if there is anything I can change or add in order to get this PR merged.

### DIFF
--- a/lib/haml_lint/reporter/github_reporter.rb
+++ b/lib/haml_lint/reporter/github_reporter.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: false
+
+module HamlLint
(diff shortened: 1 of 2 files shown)


sds (approved):
Thank you!

PR #494, opened 2024-02-19
Feature: introduce RuboCop performance

tagliala:
Hi,

I'm proposing to add RuboCop performance to avoid simple mistakes that can improve performance, like capturing matchdata when checking regexp instead of using `match?`


sds:
Generally supportive. Can we get the rest of the suite passing? Thank you.

Appraisals
```diff
@@ -32,5 +32,5 @@ appraise 'haml6.3' do
 end
 
 appraise 'rubocop1.0' do
-  gem 'rubocop', '~> 1.0.0'
+  gem 'rubocop', '~> 1.0'
```
tagliala, line 35: all green, but I had to change this

tagliala:
Probably it would be better to drop support for old RuboCop versions before

sds:
Thanks!

PR #505, opened 2024-04-24
Fix `block_keyword` method returning keywords used as arguments of method call

Adrian-Hirt:
Hey there

I noticed a bug when using one of the keywords used by `HamlLint::RubyExtraction::ChunkExtractor.block_keyword` is used as a keyword argument, e.g. for a helper method (i.e. as a symbol).

The following works without any issues:

```ruby
= helper foo: true, bar: true, if: false
```

as well as

```ruby
= helper foo: true,
         bar: true, if: false
```

However, the following code will produce an exception:

```ruby
= helper foo: true, bar: true,
         if: false
```

This is also the case for other keywords, e.g. `unless`:

```ruby
= helper foo: true, bar: true,
         unless: User.first
```

Whether using such Ruby keywords as keyword arguments is a sensible idea is another question, however the last 2 examples will produce the following exception:

```bash
RuboCop: Couldn't process the file for linting. RuntimeError: Line <lineno> should be followed by indentation. This might actually work in Haml, but it's almost a bug that it does. haml-lint cannot process.
```

The code works perfectly fine in HAML, so I guess it should also be accepted by haml-lint.

I adapted the `HamlLint::RubyExtraction::ChunkExtractor.block_keyword` method to ignore such "keyword argument" occurences of keywords, and added some testcases. Feel free to suggest edits, otherwise thanks for looking at this PR.

Have a nice day

### DIFF
--- a/lib/haml_lint/ruby_extraction/chunk_extractor.rb
+++ b/lib/haml_lint/ruby_extraction/chunk_extractor.rb
@@ -17,6 +17,10 @@ module HamlLint::RubyExtraction
                              ::Haml::Parser.new('', {})
                            end
 
+    # HAML strips newlines when handling multi-line statements (using pipes or trailing comma)
+    # We don't. So the regex must be fixed to correctly detect the start of the string.
+    BLOCK_KEYWORD_REGEX = Regexp.new(Haml::Parser::BLOCK_KEYWORD_REGEX.source.sub('^', '\A'))
+
     def initialize(document, script_output_prefix:)
       @document = document
       @script_output_prefix = script_output_prefix
@@ -692,17 +696,8 @@ module HamlLint::RubyExtraction
         return keyword
       end
 
-      code.scan(Haml::Parser::BLOCK_KEYWORD_REGEX) do |_c|
-        # Check that the keyword is actually a keyword, and not
-        # an argument to a method.
-        next_char = code[Regexp.last_match.offset(0)[1]]
-
-        next if next_char == ':'
-
-        return Regexp.last_match[1] || Regexp.last_match[2]
-      end
-
-      nil
+      return unless keyword = code.scan(BLOCK_KEYWORD_REGEX)[0]
+      keyword[0] || keyword[1]
     end
   end
 end
--- a/lib/haml_lint/ruby_extraction/chunk_extractor.rb
+++ b/lib/haml_lint/ruby_extraction/chunk_extractor.rb
@@ -692,7 +692,7 @@ module HamlLint::RubyExtraction
         return keyword
       end
 
-      code.scan(Haml::Parser::BLOCK_KEYWORD_REGEX) do |c|
+      code.scan(Haml::Parser::BLOCK_KEYWORD_REGEX) do |_c|
         # Check that the keyword is actually a keyword, and not
         # an argument to a method.
         next_char = code[Regexp.last_match.offset(0)[1]]
@@ -702,7 +702,7 @@ module HamlLint::RubyExtraction
         return Regexp.last_match[1] || Regexp.last_match[2]
       end
 
-      return nil
+      nil
     end
   end
 end
--- a/lib/haml_lint/ruby_extraction/chunk_extractor.rb
+++ b/lib/haml_lint/ruby_extraction/chunk_extractor.rb
@@ -692,8 +692,17 @@ module HamlLint::RubyExtraction
         return keyword
       end
 
-      return unless keyword = code.scan(Haml::Parser::BLOCK_KEYWORD_REGEX)[0]
-      keyword[0] || keyword[1]
+      code.scan(Haml::Parser::BLOCK_KEYWORD_REGEX) do |c|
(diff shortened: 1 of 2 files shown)


MaxLap:
Hello, thanks for the PR. I'll look into it soon.

From a quick glance, I worry that your change would break with this not idiomatic code (I didn't verify that guess):

```
- if:sym == myvar
  hello
```

I know I tried to reproduce what HAML does a lot. So considering the code literally use HAML's regex, my feeling is that a better fix would probably be elsewhere.
My gut feeling is that whatever check is happening should maybe instead be skipped when handling line continuation based on comma.

Adrian-Hirt:
Hey @MaxLap , thanks for the quick reply!

> From a quick glance, I worry that your change would break with this not idiomatic code (I didn't verify that guess):
> 
> ```
> - if:sym == myvar
>   hello
> ```

I quickly tested it, however it fails with and without my changes applied with `Lint/Syntax: unexpected token kIF_MOD`, while it works as an `if` condition when rendered with HAML, so it seems like there is another issue with the code at hand, but yes, my code indeed refuses the `if`, while the original code returns the `if`.

> I know I tried to reproduce what HAML does a lot. So considering the code literally use HAML's regex, my feeling is that a better fix would probably be elsewhere. My gut feeling is that whatever check is happening should maybe instead be skipped when handling line continuation based on comma.

Probably yes, if you have a better working solution I'd be very happy, the fix I came up with was just based on the section of haml-lint I looked at.

If I understand correctly (and please correct me if I'm wrong), the issue at hand stems from the fact that the method `HamlLint::RubyExtraction::ChunkExtractor.start_block_keyword?` returns `true` for the failing testcase, even though it should return `false` as the `if` is not starting a block.
However, for a keyword to start a block, it must be the first word in a line, right? And as such we could simply check whether the extracted keyword is the first word in the code, e.g. by changing the code of `HamlLint::RubyExtraction::ChunkExtractor.start_block_keyword?` from

```ruby
def self.start_block_keyword?(code)
  START_BLOCK_KEYWORDS.include?(block_keyword(code))
end
```

to this version:

```ruby
def self.start_block_keyword?(code)
  found_keyword = block_keyword(code)

  return false if found_keyword.nil?

  START_BLOCK_KEYWORDS.include?(found_keyword) && code.split.first == found_keyword
end
```

Would that work for you? I can update the PR if that would be okay with you, or if you have another idea I'd be happy as well :)

I wish you a pleasant day!

MaxLap:
> However, for a keyword to start a block, it must be the first word in a line, right? And as such we could simply check whether the extracted keyword is the first word in the code, e.g. by changing the code of

I would avoid using "line" in a context where sometimes, multiple lines are treated as one, as the meaning gets ambiguous. We could use "statement", or "chunk" as I ended up doing in the linter.

I've finally gotten to look into this.

As I mentioned, most of this logic had to be copied from HAML. So things not working on our side only means either an error ins such copying, or a change in an assumption due to other changes in Haml-Lint. My priority to correctly fix such a problem is to be certain of the cause.

Here's what's going on. When HAML uses a comma to treat two lines of code as one. It joins the pieces of code with a simple space. The newline is gone. This was a problem for Haml-Lint, which needs the newline to correctly represent the source code.

So Haml-Lint does find and use the actual exact source code.

As a result, For your problematic example:
```
= helper foo: true, bar: true,
         if: false
```

What HAML receives in `#block_keyword` is: `" helper foo: true, if: false, bar: true"`.

But Haml-Lint's `#block_keyword` instead receives `"helper foo: true,\n         if: false, bar: true"`.

If we look at the source of `Haml::Parser::BLOCK_KEYWORD_REGEX`, we find: `^-?\s*(?:(else|elsif|rescue|ensure|end|when)|(?:\w+(?:,\s*\w+)*\s*=\s*)?(if|begin|case|unless))\b`.

I don't know how familiar you are with Regex, but that initial `^` means the start of a line. Which in HAML's case was fine to mean the start of the chunk, since they removed the new-lines. But we have to deal with newlines, we should instead use `\A` which means start of string.

I would do this at the top of the ChunkExtractor:

```
# HAML strips newlines when handling multi-line statements (using pipes or trailing comma)
# We don't. So the regex must be fixed to correctly detect the start of the string.
BLOCK_KEYWORD_REGEX = Regexp.new(Haml::Parser::BLOCK_KEYWORD_REGEX.source.sub('^', '\A'))
```

Then using that new constant in `#block_keyword` should be all that is needed.

Could you alter your PR to do that? This way we can also have your tests which are a nice bonus.

Thank you!

Adrian-Hirt:
Hey @MaxLap , thank you for your explanation, that makes sense :)

I reverted my change to the `HamlLint::RubyExtraction::ChunkExtractor.block_keyword` method and added your regex, which handles my testcases correctly.

The tests I added are still present, as more testcoverage is always a nice thing to have :)

MaxLap:
Looks good, thank you and have a nice day.

@sds will merge it when he gets to it

PR #507, opened 2024-04-28
Fix typos

tagliala:
```
codespell **/*.rb -L filetest,compliancy,reenable -w
```

Plus a manual fix of

```
lib/haml_lint/tree/root_node.rb:8: fo ==> of, for, to, do, go
```

### DIFF
--- a/lib/haml_lint/linter/rubocop.rb
+++ b/lib/haml_lint/linter/rubocop.rb
@@ -117,7 +117,7 @@ module HamlLint
         # so the lints will be recorded then.
         @lints = []
 
-        msg = "Corrections couldn't be transfered: #{e.message} - Consider linting the file " \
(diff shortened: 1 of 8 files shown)


lib/haml_lint/linter/rubocop.rb
```diff
@@ -117,7 +117,7 @@ def transfer_corrections(coordinator, new_ruby_code)
         # so the lints will be recorded then.
         @lints = []
 
-        msg = "Corrections couldn't be transfered: #{e.message} - Consider linting the file " \
+        msg = "Corrections couldn't be transferred: #{e.message} - Consider linting the file " \
```
tagliala, line 120: Note 1: this may be considered a breaking change
Note 2: as far as I understand, this is the only correct version in both British English and American English

sds:
Thanks!

PR #535, opened 2024-12-31
Update README to show that we run Appraisal on haml 5/6 not 4/5

maddieholtzer:
**WHAT**

Update README to show that we run Appraisal on haml 5/6 not 4/5

**WHY**

Haml 4 compatibility was dropped in [this pr](https://github.com/sds/haml-lint/pull/466) as it was in `Appraisals`. There was not a corresponding update in the README to reflect this change and it still reads `we use Appraisal to test against both HAML 4 and 5` which is now out of date.


sds:
Thanks!

Issue #406, opened 2023-06-28
False positives with `Style/DisableCopsWithinSourceCodeDirective`.


mockdeep:
When upgrading to version 0.46.0, we're seeing a lot of false positives with the `Style/DisableCopsWithinSourceCodeDirective` linter. The following line triggers it for some reason:

```haml
%li.tab-navigation__tab{ **tab_attributes }
```

MaxLap:
I'm learning about a new Cop I didn't know about...

We use the `begin # rubocop:disable Style/RedundantBegin,Lint/RedundantCopDisableDirective` when we encounter tags to add valid indentation to the Ruby code send to Rubocop. (Using a block would cause scoping issues with variables)

I made a PR (#412) to fix this by instead adding a `ensure` to the begin and remove the disables. This should fix your issue.

In the meantime, I can only suggest to disable DisableCopsWithinSourceCodeDirective or to add `AllowedCops` for Style/RedundantBegin and Lint/RedundantCopDisableDirective

MaxLap:
Haml-Lint 0.47.0 has been released which fixes this. This can be closed

PR #536, opened 2024-12-31
Set `Layout/First_#{}_LineBreak` lints as disabled in rubocop config

maddieholtzer:
**What**

Set 3 of the 4 `Layout/First_#{}_LineBreak` lints as `enabled: false` in `config/forced_rubocop_config.yml`

- [Layout/FirstArrayElementLineBreak](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirstarrayelementlinebreak)
- [Layout/FirstHashElementLineBreak](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirsthashelementlinebreak)
- [Layout/FirstMethodArgumentLineBreak](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirstmethodargumentlinebreak)

Leaving 1 of the 4 as is:

- [Layout/FirstMethodParameterLineBreak](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirstmethodparameterlinebreak)

**Why**

These lints run into a conflict with [MultilinePipe](https://github.com/sds/haml-lint/blob/main/lib/haml_lint/linter/README.md#multilinepipe)

One example of this is that the autocorrect changes:

```
= link_to('View button styleguide',
  style_path('button'),
  target: '_blank',
  rel: 'noopener noreferrer')
```

to:

```
= link_to( |
  'View button styleguide', |
  style_path('button'), |
  target: '_blank', |
  rel: 'noopener noreferrer', |
  ) |
```

It is quite possible that this is showcasing a lack of understanding I have of haml syntax/best practice but I wasn't able to figure out a workaround that can satisfy both lints.

I left out [Layout/FirstMethodParameterLineBreak](https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirstmethodparameterlinebreak) because in order to trigger this (which I couldn't figure out how to do without triggering a syntax error) you are probably doing something you shouldn't in the first place (defining a method inside a haml file is something that comes across to me as [bad practice](https://stackoverflow.com/questions/52117429/haml-how-to-define-a-method)) and an additional safeguard against this is probably a good thing.

### DIFF
--- a/config/forced_rubocop_config.yml
+++ b/config/forced_rubocop_config.yml
@@ -45,6 +45,18 @@ Layout/EndAlignment:
 Layout/EndOfLine:
   Enabled: false
 
+# Complying with this lint requires the use of pipes which causes a conflict with MultilinePipe.
+Layout/FirstArrayElementLineBreak:
+  Enabled: false
+
+# Complying with this lint requires the use of pipes which causes a conflict with MultilinePipe.
+Layout/FirstHashElementLineBreak:
+  Enabled: false
+
+# Complying with this lint requires the use of pipes which causes a conflict with MultilinePipe.
+Layout/FirstMethodArgumentLineBreak:
+  Enabled: false
+
 # Turning this cop on can turn
 # = content_tag(:span) do
 #   - foo


sds:
Going to defer to @MaxLap on whether this is the right approach, as the `forced_rubocop_config.yml` was introduced by them.

MaxLap:
Those basically go against what we would consider good haml syntax. So I think it make sense to add this.

maddieholtzer:
I've thought through this a tiny bit more and realized that this doesn't apply within `:ruby` blocks where you can satisfy these lints without running into any issues with `MultilinePipe` or syntax issues. I'm not sure if that really changes things here though.

MaxLap:
I've tried using rubocop:disable comments in the past. It caused issues for users that use Style/DisableCopsWithinSourceCodeDirective, see https://github.com/sds/haml-lint/issues/406. So that's not an option.

I don't see a clean way to handle both styling at the same time. So I guess I would disable the cops. The main issue is if people used to have it active, they will see lots of changes, which is annoying. But I think those changes would be for the better?

maddieholtzer:
That is also a concern for me. I'd be curious to know how/if people really have these active in Haml right now unless they've disabled `MultilinePipe` or their Haml files are basically just entirely `:ruby` blocks. I could very well be missing something, though.

PR #537, opened 2025-01-03
Enable rubocop rules this project is compliant with

maddieholtzer:
**What**

Enable rubocop rules this project is compliant with

- `Bundler/OrderedGems`
- `Style/CollectionMethods`
- `Style/FormatString`
- `Naming/FileName`
- `Style/SingleLineBlockParams`
- `Style/TrivialAccessors`

**Why**

Since we are compliant with these rules it feels a bit weird to have them disabled in the `rubocop.yml` as we either are following the expectations intentionally and thus probably good to lock down, or the lints aren't ones that are relevant to this project if they have no issues a decade in.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,14 +11,8 @@ AllCops:
 Lint/AssignmentInCondition:
   Enabled: false
 
-Bundler/OrderedGems:
-  Enabled: false
-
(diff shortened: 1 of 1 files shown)


sds:
Thanks

PR #542, opened 2025-01-16
Re-run `appraisal install` to sync gemfiles/* versions

mjankowski:
I think what happens here is the dependabot PRs update just the main Gemfile, and that does not bring along the gemfiles/* with it? Just re-ran `bundle exec appraisal install` here to get them pinned similarly to main Gemfile.


sds:
Thanks!

PR #545, opened 2025-01-30
Update InstanceVariables linter to support :ruby filters

jdelStrother:
We sometimes use the :ruby filter as a preamble to our haml files, to set up local variables.

```haml
:ruby
  title = @page_title
%h1= title
```

How about something like this so that the InstanceVariables linter can detect ivars in filters?


sds:
Thanks!

PR #553, opened 2025-02-23
Add `StrictLocals` linter

### LINKED ISSUE #486: Feature Suggestion: require strict locals
We've been incorporating the new ["strict locals" feature in Rails 7.1](https://edgeguides.rubyonrails.org/7_1_release_notes.html#allow-templates-to-set-strict-locals) into our views. It would be great to have a haml linter to require this. Right now we're only doing it in partials, not top level templates, so ideally it would allow some configuration like [the `InstanceVariables` linter](https://github.com/sds/haml-lint/blob/main/lib/haml_lint/linter/README.md#instancevariables) to allow linting only partials files. In some projects I would lint all, though, as they pass locals from the controllers as well.

wilfison:
## StrictLocals

Checks for the presence of a `locals` magic comment at the beginning of a file. Read more about this in the [Ruby on Rails Guides](https://guides.rubyonrails.org/action_view_overview.html#strict-locals) on why it is good practice to define local variables at the beginning of the file.

**Bad:**
```haml
%h1= title
```

**Good**
```haml
-# locals: (title:)

%h1= title
```

The new linter follows the same settings as `InstanceVariables` as designed by @mockdeep  in #486 

```yaml
  StrictLocals:
    enabled: true
    file_types: partials
    matchers:
      all: .*
      partials: \A_.*\.haml\z
```

Close sds/haml-lint#486

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -103,7 +103,7 @@ linters:
     style: space
 
   StrictLocals:
-    enabled: true
+    enabled: false
     file_types: partials
     matchers:
       all: .*
--- a/config/default.yml
+++ b/config/default.yml
@@ -102,6 +102,13 @@ linters:
     enabled: true
     style: space
 
+  StrictLocals:
+    enabled: true
+    file_types: partials
+    matchers:
+      all: .*
+      partials: \A_.*\.haml\z
+
   TagName:
     enabled: true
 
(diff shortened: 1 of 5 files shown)


config/default.yml
```diff
@@ -102,6 +102,13 @@ linters:
     enabled: true
     style: space
 
+  StrictLocals:
+    enabled: true
```
sds, line 106: I would suggest disabling this by default, since it seems like a relatively invasive change for existing codebases.

sds (commented):
Looks great—thank you. Will adjust the default to be disabled so this doesn't come as a surprise to existing users.

PR #561, opened 2025-03-20
Allow `StrictLocals` to be disabled by a comment

davidcornu:
Follow-up to https://github.com/sds/haml-lint/pull/553.

Because `StrictLocals` operates on the root node, it ignores the potential presence of a `haml-lint: disable` comment on the first line which means it can't be disabled that way.

This is unfortunate as I've often found those comments to be a useful tool to allow gradual adoption of a new lint.


davidcornu:
That was quick. Thanks @sds ❤️

sds:
Released in [0.61.1](https://github.com/sds/haml-lint/releases/tag/v0.61.1)

PR #562, opened 2025-03-21
Allow the strict locals comment to be preceeded by other comments

davidcornu:
Follow-up to https://github.com/sds/haml-lint/pull/561 and https://github.com/sds/haml-lint/pull/553

Certain linters like `ViewLength` can be disabled via a `haml-lint:disable` comment but it must be the first thing in the file https://github.com/sds/haml-lint/blob/b248f8d3dae932fb2d285bbc2e441d52804b2a27/lib/haml_lint/linter/view_length.rb#L15-L17

This clashes with the `StrictLocals` lint as it currently reports an error for cases like this

```haml
-# haml-lint:disable ViewLength
-# locals: (foo: )
```

as it expects the strict locals comment to come first despite it being completely valid from Rails' standpoint. As per https://guides.rubyonrails.org/action_view_overview.html#strict-locals it can be anywhere in the file:

> Action View will process the `locals:` signature in any templating engine that supports `#`-prefixed comments, and will read the signature from any line in the partial.

As a middle ground I've 
- Kept the restriction that `StrictLocals` can only be disabled by comment if it's the first thing in the file (to mirror `ViewLength`'s behaviour)
- Allowed the strict locals comment to appear anywhere in a series of comments at the top of the file (as I still think having it at the top is preferable)

### DIFF
--- a/lib/haml_lint/linter/strict_locals.rb
+++ b/lib/haml_lint/linter/strict_locals.rb
@@ -14,12 +14,16 @@ module HamlLint
     def visit_root(root)
       return unless enabled?(root)
 
-      first_child = root.children.first
-      return if first_child.is_a?(HamlLint::Tree::HamlCommentNode) &&
-                    first_child.is_strict_locals?
+      # Rails technically allows the comment to be anywhere in the file,
+      # but as a best practice they should be at the top of the file.
+      # https://guides.rubyonrails.org/action_view_overview.html#strict-locals
+      # https://github.com/rails/rails/blob/v8.0.2/actionview/lib/action_view/template.rb#L368
+      found =
+        root.children
+            .take_while { |child| child.is_a?(HamlLint::Tree::HamlCommentNode) }
+            .any?(&:is_strict_locals?)
 
-      # Check whether this linter is disabled by a comment
-      return if first_child.disabled?(self)
+      return if found
 
       record_lint(DummyNode.new(1), failure_message)
     end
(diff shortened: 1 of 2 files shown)


davidcornu:
cc @wilfison - I think the compromise is acceptable but would appreciate your opinion here

wilfison:
@davidcornu 

I understand what you did here and I think you are right about allowing comments everywhere.

Do you think it would be interesting to add an additional configuration to force comments at the beginning? Since it is a good practice.

It would be something like:

```yaml
# config/default.yml

  StrictLocals:
    enabled: true
    EnforcedStyle: top # or everywhere
```

Something like that

davidcornu:
> I understand what you did here and I think you are right about allowing comments everywhere.
> 
> Do you think it would be interesting to add an additional configuration to force comments at the beginning? Since it is a good practice.

@wilfison Ah so I'm not allowing comments _everywhere_. My change makes it so that instead of requiring the comment to be the very first node in the document, it can now be one of many comments at the beginning at the doc.

So basically, where you could only do this before

```haml
-# locals: ()
%h1 Hello
```

you can now do

```haml
-# haml-lint:ignore ViewLength
-# Some other comment
-# locals: ()
%h1 Hello
```

but not

```haml
%h1 Hello
-# locals: ()
```

or even

```haml
-# haml-lint:ignore ViewLength

-# locals: ()
%h1 Hello
```

(see https://github.com/sds/haml-lint/pull/562/files#diff-c0a289343f40c507ddb7cbab1d87e0f983ebfcb7a6b38f9ff55f48ee8f7534edR21-R24)

because I definitely agree that the best practice should be to put comments at the beginning (but maybe not the very beginning because of the aforementioned clashes with `haml-lint:ignore` comments.

I could make it configurable so that comments can be truly anywhere but I would question whether that's an advisable pattern in the first place.

davidcornu:
@wilfison no worries! Are there any code comments I could improve to make it clearer?

davidcornu:
Thanks @sds !

PR #573, opened 2025-06-22
Faster runs via RuboCop::Runner

jdelStrother:
Every time we call `RuboCop::CLI#run` (which happens once per file being linted), it sets up a new RuboCop::Runner.
This is pretty slow, especially from `RuboCop::Options#parse` and `RuboCop::Runner#mobilized_cop_classes`.

We can skip all this repeated setup by using RuboCop::Runner directly. This also means we don't have to capture the stdout/stderr streams from CLI, since the relevant information is available directly on the runner instance.

Before:

```
626 files inspected, 0 lints detected

________________________________________________________
Executed in   23.22 secs    fish           external
   usr time   20.12 secs    0.32 millis   20.12 secs
   sys time    2.57 secs    2.49 millis    2.56 secs
```

After:

```
626 files inspected, 0 lints detected

________________________________________________________
Executed in   17.64 secs    fish           external
   usr time   15.30 secs    0.34 millis   15.30 secs
   sys time    1.93 secs    2.60 millis    1.93 secs
```

I've introduced a subclass of RuboCop::Runner to handle some of the customizations we need. ruby-lsp have a similar approach here - https://github.com/Shopify/ruby-lsp/blob/13bb8b0addf96248a951dca988e40127c86b6ba8/lib/ruby_lsp/requests/support/rubocop_runner.rb#L54, which might be useful for comparison.

### DIFF
--- a/lib/haml_lint/linter/rubocop.rb
+++ b/lib/haml_lint/linter/rubocop.rb
@@ -98,9 +98,9 @@ module HamlLint
 
     def rubocop_config_for(path)
       user_config_path = ENV['HAML_LINT_RUBOCOP_CONF'] || config['config_file']
-      user_config_path ||= self.class.rubocop_config_store.user_rubocop_config_path_for(path)
+      user_config_path ||= rubocop_config_store.user_rubocop_config_path_for(path)
       user_config_path = File.absolute_path(user_config_path)
-      self.class.rubocop_config_store.config_object_pointing_to(user_config_path)
+      rubocop_config_store.config_object_pointing_to(user_config_path)
     end
 
     # Extracted here so that tests can stub this to always return true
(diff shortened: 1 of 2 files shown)


jdelStrother:
I've also got a [POC here](https://github.com/jdelStrother/haml-lint/commit/rubocop-cache) that makes use of RuboCop's cache, which dramatically improves times on already-processed files:

```
626 files inspected, 0 lints detected

________________________________________________________
Executed in    5.22 secs    fish           external
   usr time    4.30 secs    0.32 millis    4.30 secs
   sys time    0.88 secs    4.40 millis    0.87 secs
```

RuboCop doesn't use its cache on code from stdin, so when we generate ruby code from the haml file for RuboCop to lint, we could write it to, eg, `tmp/haml-lint/app/views/posts/index.html.haml.rb`.
Relying on a tmp directory doesn't seem great though, so maybe we can defer that idea to a future PR.

sds:
Amazing work! Thank you

PR #576, opened 2025-06-27
Enable RuboCop caching

jdelStrother:
Because we call RuboCop::Runner with the extracted ruby code on stdin, RuboCop thinks that its cache ought to be disabled. (https://github.com/rubocop/rubocop/blob/0bb3b4ffdf0568ea978512947e9a027a35caba83/lib/rubocop/runner.rb#L253-L263)

Override the `RuboCop::Runner#cached_run?` method so that if haml-lint itself isn't using code from stdin, RuboCop is allowed to cache results based on a hash of the haml file.

Before:

```
$ bundle exec haml-lint app/{components,views}

626 files inspected, 0 lints detected

________________________________________________________
Executed in    5.06 secs    fish           external
   usr time   23.90 secs  340.00 micros   23.90 secs
   sys time    3.99 secs  993.00 micros    3.99 secs
```

After:

```
$ bundle exec haml-lint app/{components,views}

626 files inspected, 0 lints detected

________________________________________________________
Executed in    2.65 secs    fish           external
   usr time    8.78 secs    0.53 millis    8.78 secs
   sys time    2.19 secs    5.11 millis    2.18 secs
```

---

This seems to have been working well for me locally, but would definitely appreciate some more testing from others.

### DIFF
--- a/lib/haml_lint/document.rb
+++ b/lib/haml_lint/document.rb
@@ -14,6 +14,9 @@ module HamlLint
     # @return [String] Haml template file path
     attr_reader :file
 
+    # @return [Boolean] true if source was read directly from `file` on-disk (rather than from stdin)
+    attr_reader :file_on_disk
+
     # @return [Boolean] true if source changes (from autocorrect) should be written to stdout instead of disk
     attr_reader :write_to_stdout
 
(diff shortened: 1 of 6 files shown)


sds:
Released in [0.64.0](https://github.com/sds/haml-lint/releases/tag/v0.64.0). Thanks!

PR #581, opened 2025-07-18
Alphabetize lints when generating todo file

maddieholtzer:
**What**

Alphabetize lints when generating todo file

**Why**

Makes the todo files a bit easier to read through.

### DIFF
--- a/lib/haml_lint/reporter/disabled_config_reporter.rb
+++ b/lib/haml_lint/reporter/disabled_config_reporter.rb
@@ -88,7 +88,7 @@ module HamlLint
       output = []
       output << HEADING
       output << 'linters:' if linters_with_lints.any?
-      linters_with_lints.each do |linter, files|
+      linters_with_lints.sort.each do |linter, files|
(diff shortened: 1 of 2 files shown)


lib/haml_lint/reporter/disabled_config_reporter.rb
```diff
@@ -88,7 +88,7 @@ def config_file_contents
       output = []
       output << HEADING
       output << 'linters:' if linters_with_lints.any?
-      linters_with_lints.each do |linter, files|
+      linters_with_lints.sort.each do |linter, files|
```
maddieholtzer, line 91: `linters_with_lints` is a hash with lints and the files they've failed in that is slowly built up as defined here: https://github.com/sds/haml-lint/blob/c5551710006f7921bf2a5b79f442356f20f13130/lib/haml_lint/reporter/disabled_config_reporter.rb#L71-L80

This `.sort` call changes this from a hash to a 2d array which appears to still work in `generate_config_for_linter` so doesn't seem necessary to convert back to a hash.

PR #582, opened 2025-07-18
Test against ruby 3.4

maddieholtzer:
**What**

Test against ruby 3.4/use ruby 3.4 in overcommit

**Why**

Ruby 3.4 has been out for half a year now

**Notes**

The last update to 3.3 was a year and a half ago in [this pr](https://github.com/sds/haml-lint/pull/472)


maddieholtzer:
Looks like this breaks some stuff when testing against ruby 1 :(

PR #584, opened 2025-07-23
Make sure todo file generates with a newline at the end

maddieholtzer:
**What**

Make sure todo file generates with a newline at the end

**Why**

Everytime I regenerate a haml lint todo file I have to manually add in a newline so that we don't lose it.

**Notes**

This matches how rubocop works as can be seen in the contents of [spec/rubocop/formatter/disabled_config_formatter_spec.rb](https://github.com/rubocop/rubocop/blob/6c522f8db358fdcc19381cd88e77f0d08ffed1b6/spec/rubocop/formatter/disabled_config_formatter_spec.rb#L105-L121):

```
let(:expected_rubocop_todo) do
  [heading,
    '# Offense count: 2',
    'Test/Cop1:',
    '  Exclude:',
    "    - 'Gemfile'",
    "    - 'test_a.rb'",
    "    - 'test_b.rb'",
    '',
    '# Offense count: 1',
    'Test/Cop2:',
    '  Exclude:',
    "    - '**/*.blah'",
    '    - !ruby/regexp /.*/bar/*/foo\.rb$/',
    "    - 'test_a.rb'",
    ''].join("\n")
end
```

### DIFF
--- a/spec/haml_lint/reporter/disabled_config_reporter_spec.rb
+++ b/spec/haml_lint/reporter/disabled_config_reporter_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe HamlLint::Reporter::DisabledConfigReporter do
 
       it 'creates a file with just a header' do
         subject
-        config.should == described_class::HEADING
+        config.should == described_class::HEADING + "\n"
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/haml_lint/reporter/disabled_config_reporter_spec.rb
```diff
@@ -111,6 +111,7 @@
             '    exclude:',
             '      - "other-filename.haml"',
             '      - "some-filename.haml"',
+            '',
           ].join("\n")
```
maddieholtzer, line 115: What this actually looks like to the end user (at least us) is whether or not one of these bad boys shows up in the diff:

Screenshot 2025-07-23 at 17 28 52

PR #597, opened 2025-08-28
CI: add Ruby 3.4 to build matrix

olleolleolle:
This PR only updates the CI matrix to the latest generally available Ruby versions.

Side-note: Ruby 3.1 is EOL since a few months, no more support. https://endoflife.date/ruby


olleolleolle:
Yes, we can add a dependency on the base64 gem, which from Ruby 3.4 isn't shipped with Ruby itself.

sds:
Thank you!

PR #610, opened 2025-11-09
Add metadata for e.g. Renovate

Nymuxyzo:


Nymuxyzo:
I don't know if it's worth going into more detail here. I just wanted to add the metadata for Renovate, and then Rubocop complained because one rule had been renamed. Furthermore, I fixed that, but now even more things are going haywire because there's this old Appraisal config with Rubocop 1.0.

sds:
Yeah, I'm not sure what's going on here. FWIW [all tests pass on the latest `main` branch](https://github.com/sds/haml-lint/actions/runs/19048053737/job/54401131543), and a [dummy PR with no code changes also passes](https://github.com/sds/haml-lint/actions/runs/19306588258/job/55215595152?pr=611). so it does appear like something is off with this change?

If you are able to figure out why, happy to merge the change. Otherwise will close.

PR #641, opened 2026-04-08
Fix RuboCop deprecation warnings

mattmenefee:
## Summary

- Switch `require: rubocop-performance` to `plugins: rubocop-performance` per RuboCop's new plugin system
- Rename `Naming/PredicateName` cop to `Naming/PredicatePrefix` to match the upstream rename

## Test plan

- [x] `bundle exec rubocop` runs clean with no warnings and no offenses

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,4 @@
-require:
+plugins:
   - rubocop-performance
 
 AllCops:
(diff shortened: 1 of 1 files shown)


mattmenefee:
Closing — these config changes break backward compatibility with RuboCop 1.0.x, which is still in the supported version range (gemspec requires >= 1.0). The old names produce harmless deprecation warnings in newer RuboCop but the new names cause hard errors in RuboCop 1.0.

PR #656, opened 2026-06-20
Add `UnescapedHtml` linter

### LINKED ISSUE #550: Linter suggestion - Avoid Unescape HTML
Haml contains a method to ensure that generated html is not escaped. Specifically, you can use `!=` instead of `=` if you want to avoid escaping.

https://haml.info/docs/yardoc/file.REFERENCE.html#unescaping_html

```
= "I feel <strong>!"
!= "I feel <strong>!"
```

However, like using `raw` and `h()` and `.html_safe` and friends, it's very easy to accidentally create security holes in your application

```
!= "Username: <strong>#{user.name}</strong>"
```

In most cases `!=` is undesirable, and can be avoided by using safer methods (e.g. in Rails, by using SafeBuffers when building in views or helpers).

I'd therefore like to see a linter that can flag up uses of `!=`, in case they slip through code review.

wilfison:
Closes #550

## Summary

Adds a new `UnescapedHtml` linter that flags HAML's unescaped-output markers, which bypass HTML escaping. As noted in #550, just like `raw`, `html_safe`, and `h()` in Rails, these markers make it easy to accidentally introduce XSS holes when the output includes user-controlled data — for example:

```haml
!= "Username: <strong>#{user.name}</strong>"
```

RuboCop's `Rails/OutputSafety` cop catches `raw`/`html_safe`/`safe_concat`, but it never sees these because they are a HAML construct, not Ruby. This linter fills that gap.

## What is flagged

The linter targets every HAML marker that emits **unescaped** output of **dynamic** content:

| HAML              | Flagged? | Reason                                         |
| ----------------- | :------: | ---------------------------------------------- |
| `= value`         |    no    | escaped output                                 |
| `~ value`         |    no    | escaped, whitespace-preserved output           |
| `!= value`        | **yes**  | unescaped output                               |
| `!~ value`        | **yes**  | unescaped, whitespace-preserved output         |
| `! text #{value}` | **yes**  | unescaped dynamic plain text                   |
| `%tag!= value`    | **yes**  | unescaped tag content                          |
| `%tag!~ value`    | **yes**  | unescaped, preserved tag content               |
| `! static text`   |    no    | static plain text — no injection vector        |
| `%p= a != b`      |    no    | the Ruby `!=` operator, not an unescape marker |

The distinction is intentional: only unescaped output of _dynamic_ content is reported, since static markup carries no injection risk.

## Implementation notes

- Detection is based on the node's source marker, not `@value[:escape_html]`:
  haml-lint parses with HTML escaping disabled, so both `=` and `!=` report
  `escape_html: false` and cannot be told apart that way.
- For `:script` nodes the check is anchored at the start of the source
  (`/\A\s*!/`), so it covers `!=`, `!~`, and the dynamic plain-text `!` form
  while never matching a Ruby `!=` operator inside an escaped expression.
- For `:tag` nodes a new `TagNode#unescape_html?` helper inspects the content
  marker that follows the tag name and attributes. The existing attribute-source
  scan was refactored into a single memoized `parsed_attributes_source` that now
  also exposes the trailing `inline_marker_source`, keeping tag-source parsing in
  one place.
- No autocorrect: rewriting `!=` to `=` would change behavior and could escape
  HTML the author intentionally left raw, so this is detection only.

## Enabled by default

The linter ships enabled in `config/default.yml`.

## Testing

```bash
bundle exec rspec spec/haml_lint/linter/unescaped_html_spec.rb spec/haml_lint/tree
bundle exec rubocop
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # HAML-Lint Changelog
 
+### Unreleased
+
+* Add `UnescapedHtml` linter to flag use of `!=`/`!~`/`!` unescaped HTML output
+
 ### 0.73.0
 
 * Relax `parallel` dependency from `~> 1.10` to `>= 1.10`
--- a/config/default.yml
+++ b/config/default.yml
@@ -118,6 +118,9 @@ linters:
   TrailingWhitespace:
     enabled: true
 
+  UnescapedHtml:
+    enabled: true
+
   UnnecessaryInterpolation:
     enabled: true
 
--- a/lib/haml_lint/linter/README.md
+++ b/lib/haml_lint/linter/README.md
@@ -31,6 +31,7 @@ Below is a list of linters supported by `haml-lint`, ordered alphabetically.
 * [TagName](#tagname)
 * [TrailingEmptyLines](#trailingemptylines)
 * [TrailingWhitespace](#trailingwhitespace)
+* [UnescapedHtml](#unescapedhtml)
 * [UnnecessaryInterpolation](#unnecessaryinterpolation)
 * [UnnecessaryStringOutput](#unnecessarystringoutput)
 * [ViewLength](#viewlength)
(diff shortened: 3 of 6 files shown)


gravitystorm:
Thanks @wilfison ! This is great, I had no idea how many different ways there are in haml to output unescaped text. 

This new linter has already flagged up one example in our codebase that we'd overlooked, despite fixing what we thought was everything before I created the original issue!

